PR #22, opened 2023-05-23
Update CONTRIBUTING.md with license bit

### LINKED ISSUE #2: Licence for tool source
SPDX does not offer a general "public domain" declaration, and the concept of "public domain"
is legally problematic.  See https://wiki.spdx.org/view/Legal_Team/Decisions/Dealing_with_Public_Domain_within_SPDX_Files
for an overview of the issues.

Therefore we should consider choosing a copyright licence for the tool source code.  Most or all of our
current dependencies use BSD-3-Clause.

The advisories themselves can remain public domain.

david-christiansen:
This fixes #2.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,10 @@ Feel free to do either or both of these as you see fit (we recommend you do both
    a CVE for you. In this case, you can add the GHSA ID to the advisory via the
    `aliases` field.
 
+### License
+
+All published security advisories are released under [CC0](https://creativecommons.org/share-your-work/public-domain/cc0/). By contributing an advisory, you agree to release the entire content of the advisory (including machine-readable metadata, example code, and textual descriptions) under CC0.
+
 ## Criteria
 
 This is a database of security vulnerabilities. The following are


CONTRIBUTING.md
```diff
@@ -25,6 +25,10 @@ Feel free to do either or both of these as you see fit (we recommend you do both
    a CVE for you. In this case, you can add the GHSA ID to the advisory via the
    `aliases` field.
 
+### License
+
+All published security advisories are released under [CC0](https://creativecommons.org/share-your-work/public-domain/cc0/). By contributing an advisory, you agree to release the entire content of the advisory (including machine-readable metadata, example code, and textual descriptions) under CC0.
```
blackheaven, line 30: We agreed on [BSD3](https://github.com/haskell/security-advisories/blob/main/meeting-notes/2023-05-17.md#license-for-tool-source-code)

david-christiansen: My recollection is that we agreed that the Haskell code in the `code` directory will be BSD3, while the contents of the database itself are CC0.

blackheaven: oh, sorry, yes, you're right

blackheaven (approved):

TristanCacqueray (approved):

PR #24, opened 2023-05-30
Implement check-advisories CI job

frasertweedale:
This PR:

- add the `advisories/` directory
- patches the `haskell-ci` workflow to:
  - upload the `hsec-tools` executable as an artifact
  - invoke the `check-advisories` reusable workflow
  - an update to the `REGENDATA` ensures the patch will be re-applied by `haskell-ci regenerate`
- adds the `check-advisories` reusable workflow, which
  - downloads the `hsec-tools` artifact from this CI run
  - checks files matching `advisories/*.md` as well as `EXAMPLE_ADVISORY.md`
- update build matrix, moving from GHC 9.6.1 to 9.6.2

```
450de57 (Fraser Tweedale, 24 minutes ago)
   ci: GHC 9.6.1 -> 9.6.2

   Update from GHC 9.6.1 to 9.6.2 in the build matrix.

   Note the MANUAL CHANGES to .github/haskell-ci.patch.  This is needed any
   time the referenced `matrix.compiler` is replaced or removed from the
   matrix.  Unless/until we find a better approach.

1a9fac1 (Fraser Tweedale, 33 minutes ago)
   ci: patch haskell-ci workflow to upload exe and call check-advisories

   Provide a patch to haskell-ci.yml, which haskell-ci will apply when 
   regenerating the haskell-ci workflow.  This patch adds steps to upload the
   hsec-tools executable as an artifact, and invoke the check-advisories
   reusable workflow (which was added in a previous commit).

   There is one notable caveat to this approach.  We use a conditional based
   on the `matrix.compiler` variable to ensure that we only upload one variant
   of the executable.  Currently, ghc-9.6.1.  The patch file will need to be
   manually updated as the build matrix evolves over time (for example, when
   moving from 9.6.1 to 9.6.2).

   Fixes: https://github.com/haskell/security-advisories/issues/10

b281264 (Fraser Tweedale, 2 hours ago)
   ci: add check-advisories reusable workflow

   Define a reusable workflow to checking the security advisories.  A 
   subsequent commit will patch `haskell-ci.yml` to call this workflow.

b4ecf12 (Fraser Tweedale, 12 hours ago)
   create advisories directory
```


.github/workflows/check-advisories.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: ~/.local/bin
+      - run: chmod +x ~/.local/bin/hsec-tools
+      - name: run checks
+        run: |
+          cd source
+          RESULT=0
+          for FILE in $(find advisories EXAMPLE_ADVISORY.md -type f -name "*.md") ; do
```
blackheaven, line 26: Do we intend to limit advisories names? [see for vs while](https://unix.stackexchange.com/a/24278/373072)


```suggestion
          ls advisories/*.md EXAMPLE_ADVISORY.md | while read -r FILE ; do
```

blackheaven (approved):

.github/workflows/check-advisories.yml
frasertweedale: @blackheaven thanks for pointing that out.  I've modified it, but still using `find` so that we can handle a nested structure (if we want that).

blackheaven: I think we should enforce a particular structure, otherwise we might check eventual files associated with a particular advisory.

frasertweedale: I think it's a bit early to predict what we will need, or how that could change over time.  If you're satisfied with the current behaviour, we can merge it.  We will refine or evolve it as needed.

Issue #3, opened 2023-05-16
OSV data model compatibility


frasertweedale:
Ensure that our advisory data model is compatible with the OSV schema.

We can implement export / conversion later.  For now, let's just make sure that our data type can be correctly converted to the OSV schema.

OSV schema: https://ossf.github.io/osv-schema/

TristanCacqueray:
Using `yajsv` along with the `schema.json`, we can validate the data model, for example, this data:

```yaml
id: HSEC-0000-0000
modified: 2021-01-31T00:00:00Z
```

... is valid according to:

```ShellSession
$ yajsv -s schema.json test.yaml
test.yaml: pass
```

Here is an example of how we can fit the example advisory for this data model:

```yaml
id: HSEC-0000-0000
modified: 2021-01-31T00:00:00Z
aliases: ["CVE-XXX"]
affected:
   - package:
       ecosystem: hackage
       name: package-name
       purl: https://example.com
     severity:
       - type: CVSS_V3
         score: CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H
     ranges:
       - type: SEMVER
         events:
           - introduced: "1.1.0"
           - fixed: "1.2.0.5"
```

akacase:
I'll review this tonight and tomorrow; feel free to assist, as I might overlook something @mihaimaruseac. I'll comment here on issues or if I don't find any.

TristanCacqueray:
Here is another example from the rustsec/advisory-db:

toml: https://github.com/rustsec/advisory-db/blob/main/crates/enumflags2/RUSTSEC-2023-0035.md
osv: https://github.com/rustsec/advisory-db/blob/osv/crates/RUSTSEC-2023-0035.json

mihaimaruseac:
Oh, I verified this, was planning to report on it on today's meeting but I guess that got cancelled?

I think we should be good here.

TristanCacqueray:
@mihaimaruseac we had a chat about the affected version string, which might be complicated to convert into the OSV ranges. For example:

- Bug introduced in 1.8 (stable release)
- Affect 2.0 and 2.1
- Bug discovered and fixed in 1.9 and 2.2

Should we parse `(>= 1.8 && < 1.9) || (>= 2.0 && < 2.2)`? Or should we change the schema to something closer to the desired OSV range, e.g.:

```
affects:
  - introduced: 1.8
    fixed: 1.9
  - introduced: 2.0
[...]

PR #29, opened 2023-06-07
Update affected version format

TristanCacqueray:
## hsec-tools

Related: #3

### DIFF
--- a/.github/workflows/check-advisories.yml
+++ b/.github/workflows/check-advisories.yml
@@ -23,8 +23,10 @@ jobs:
         run: |
           cd source
           RESULT=0
+          # Remove the begining of the README to extract the example.
+          (echo '```toml'; sed -e '1,/```toml/d' README.md) > EXAMPLE_README.md
           while read FILE ; do
             echo -n "$FILE: "
             hsec-tools check < "$FILE" || RESULT=1
-          done < <(find advisories EXAMPLE_ADVISORY.md -type f -name "*.md")
+          done < <(find advisories EXAMPLE_README.md EXAMPLE_ADVISORY.md -type f -name "*.md")
           exit $RESULT
--- a/README.md
+++ b/README.md
@@ -75,14 +75,15 @@ keywords = ["ssl", "mitm"]
 
 # Table of canonical paths to vulnerable declarations in the package (optional)
 # that describes which versions impacted by this advisory used that particular
-# name (e.g. if an affected function or datatype was renamed between versions). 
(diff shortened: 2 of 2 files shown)


TristanCacqueray:
That's odd the tests passed, I was expecting a parse error.

.github/workflows/check-advisories.yml
```diff
@@ -19,12 +19,12 @@ jobs:
...
+      - name: Run advisories checks
         run: |
           cd source
           RESULT=0
-          find advisories EXAMPLE_ADVISORY.md -type f -name "*.md" | while read FILE ; do
+          while read FILE ; do
             echo -n "$FILE: "
             hsec-tools check < "$FILE" || RESULT=1
-          done
+          done < <(find advisories EXAMPLE_ADVISORY.md -type f -name "*.md")
```
frasertweedale: I am curious why the old form did not result in exit status = 1 with bad advisory data, but the new form acts correctly.  Shell is weird.  Do you know what is the reason?

frasertweedale: Here's the answer: https://stackoverflow.com/questions/16854280/a-variable-modified-inside-a-while-loop-is-not-remembered/16854326#16854326

Quite the "gotcha".  Thanks for detecting and fixing this bug!

.github/workflows/check-advisories.yml
```diff
@@ -23,8 +23,10 @@ jobs:
         run: |
           cd source
           RESULT=0
+          # Remove the begining of the README to extract the example.
+          (echo '```toml'; sed -e '1,/```toml/d' README.md) > EXAMPLE_README.md
           while read FILE ; do
             echo -n "$FILE: "
             hsec-tools check < "$FILE" || RESULT=1
-          done < <(find advisories EXAMPLE_ADVISORY.md -type f -name "*.md")
+          done < <(find advisories EXAMPLE_README.md EXAMPLE_ADVISORY.md -type f -name "*.md")
```
frasertweedale: Very nice idea.

frasertweedale (approved):

frasertweedale:
@TristanCacqueray thanks for this.  If there are no more changes to be added to this PR, please go ahead and merge it.

TristanCacqueray:
I meant to add the `limit` attribute, and detect unknowns keys (the parser presently ignore them). But that can be done in a follow-up, and this PR already gets us closer to the final schema.

Thank you for the review!

PR #33, opened 2023-06-11
Add OSV data type

TristanCacqueray:
## hsec-tools

### DIFF
--- a/code/hsec-tools/src/Security/OSV.hs
+++ b/code/hsec-tools/src/Security/OSV.hs
@@ -1,0 +1,223 @@
+-- | This module contains the OSV datatype and its ToJSON instance.
+-- The module was initialized with http://json-to-haskell.chrispenner.ca/
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Security.OSV where
+
+import Data.Aeson (ToJSON(..), FromJSON(..), Value(..), (.:), (.=), object)
+import Data.Aeson.Types (prependFailure, typeMismatch)
+import Data.Text (Text)
+
+data Affected = Affected
+  { affectedRanges :: [Ranges]
+  , affectedPackage :: Package
+  , affectedEcosystemSpecific :: EcosystemSpecific
+  , affectedDatabaseSpecific :: DatabaseSpecific
+  } deriving (Show, Eq, Ord)
+
+data Affects = Affects
+  { affectsOs :: [Value]
+  , affectsArch :: [Value]
+  , affectsFunctions :: [Value]
+  } deriving (Show, Eq, Ord)
+
+data DatabaseSpecific = DatabaseSpecific
+  { databaseSpecificCvss :: Maybe Value
+  , databaseSpecificCategories :: [Value]
+  , databaseSpecificInformational :: Text
+  } deriving (Show, Eq, Ord)
+
+data EcosystemSpecific = EcosystemSpecific
+  { ecosystemSpecificAffects :: Affects
+  } deriving (Show, Eq, Ord)
+
+data Events = Events
+  { eventsIntroduced :: Text
+  } deriving (Show, Eq, Ord)
+
+data Model = Model
+  { modelDetails :: Text
+  , modelId :: Text
+  , modelSummary :: Text
+  , modelRelated :: [Value]
+  , modelAffected :: [Affected]
+  , modelAliases :: [Value]
+  , modelPublished :: Text
+  , modelReferences :: [References]
+  , modelSeverity :: [Value]
+  , modelModified :: Text
+  } deriving (Show, Eq, Ord)
+
+data Package = Package
+  { packageName :: Text
+  , packageEcosystem :: Text
+  , packagePurl :: Text
+  } deriving (Show, Eq, Ord)
+
+data Ranges = Ranges
+  { rangesEvents :: [Events]
+  , rangesType :: Text
+  } deriving (Show, Eq, Ord)
+
+data References = References
+  { referencesType :: Text
+  , referencesUrl :: Text
+  } deriving (Show, Eq, Ord)
+
+instance ToJSON Affected where
+  toJSON Affected{..} = object
+    [ "ranges" .= affectedRanges
+    , "package" .= affectedPackage
+    , "ecosystem_specific" .= affectedEcosystemSpecific
+    , "database_specific" .= affectedDatabaseSpecific
+    ]
+
+instance ToJSON Affects where
+  toJSON Affects{..} = object
+    [ "os" .= affectsOs
+    , "arch" .= affectsArch
+    , "functions" .= affectsFunctions
+    ]
+
+instance ToJSON DatabaseSpecific where
+  toJSON DatabaseSpecific{..} = object
+    [ "cvss" .= databaseSpecificCvss
+    , "categories" .= databaseSpecificCategories
+    , "informational" .= databaseSpecificInformational
+    ]
+
+instance ToJSON EcosystemSpecific where
+  toJSON EcosystemSpecific{..} = object
+    [ "affects" .= ecosystemSpecificAffects
+    ]
+
+instance ToJSON Events where
+  toJSON Events{..} = object
+    [ "introduced" .= eventsIntroduced
+    ]
+
+instance ToJSON Model where
+  toJSON Model{..} = object
+    [ "details" .= modelDetails
+    , "id" .= modelId
+    , "summary" .= modelSummary
+    , "related" .= modelRelated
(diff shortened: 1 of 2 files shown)


blackheaven (approved):

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+  { modelDetails :: Text
+  , modelId :: Text
+  , modelSummary :: Text
+  , modelRelated :: [Value]
+  , modelAffected :: [Affected]
+  , modelAliases :: [Value]
+  , modelPublished :: Text
+  , modelReferences :: [References]
+  , modelSeverity :: [Value]
+  , modelModified :: Text
```
frasertweedale, line 49: I believe all fields are optional except `id` and `modified`.

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+
+import Data.Aeson (ToJSON(..), FromJSON(..), Value(..), (.:), (.=), object)
+import Data.Aeson.Types (prependFailure, typeMismatch)
+import Data.Text (Text)
+
+data Affected = Affected
+  { affectedRanges :: [Ranges]
+  , affectedPackage :: Package
+  , affectedEcosystemSpecific :: EcosystemSpecific
+  , affectedDatabaseSpecific :: DatabaseSpecific
```
frasertweedale, line 16: There is also the `versions` field.  I'm OK if we exclude it and use only `ranges` within our tooling.  I would eventually like to extract the osv type to its own package so it can be generally useful for OSV processing.  For that we would need the fields that we (*security-advisories* as producer) don't use.  It's OK if we don't do that now - but maybe a comment is a good idea?

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+data Model = Model
+  { modelDetails :: Text
+  , modelId :: Text
+  , modelSummary :: Text
+  , modelRelated :: [Value]
+  , modelAffected :: [Affected]
+  , modelAliases :: [Value]
+  , modelPublished :: Text
+  , modelReferences :: [References]
+  , modelSeverity :: [Value]
```
frasertweedale, line 48: we should expand the severity field to a record with `type` and `score`, or even a special purpose datatype for the registered types (`CVSS_V2` and `CVSS_V3`)

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+  , modelModified :: Text
+  } deriving (Show, Eq, Ord)
+
+data Package = Package
+  { packageName :: Text
+  , packageEcosystem :: Text
+  , packagePurl :: Text
```
frasertweedale, line 55: `purl` field is optional.

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+
+data EcosystemSpecific = EcosystemSpecific
+  { ecosystemSpecificAffects :: Affects
+  } deriving (Show, Eq, Ord)
+
+data Events = Events
+  { eventsIntroduced :: Text
+  } deriving (Show, Eq, Ord)
+
+data Model = Model
```
frasertweedale, line 39: I think we are missing the `schema_version` and `credits` fields
- https://ossf.github.io/osv-schema/#schema_version-field
- https://ossf.github.io/osv-schema/#credits-fields

frasertweedale (commented):
A quick pass over the diff.  There are some missing fields and several fields that need to be made optional.

I would also prefer if as many fields as possible be refined from `Text` to more constrained, domain-specific types.  But that can be done subsequent to merging this initial work.

*edit* oh and there are some HLint nags, you probably saw it already :)

code/hsec-tools/src/Security/OSV.hs
mihaimaruseac: That is true.

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+
+data Package = Package
+  { packageName :: Text
+  , packageEcosystem :: Text
+  , packagePurl :: Text
+  } deriving (Show, Eq, Ord)
+
+data Ranges = Ranges
+  { rangesEvents :: [Events]
+  , rangesType :: Text
```
mihaimaruseac, line 60: I think we're missing `repo` here

https://ossf.github.io/osv-schema/#affectedrangesrepo-field

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+  , databaseSpecificCategories :: [Value]
+  , databaseSpecificInformational :: Text
+  } deriving (Show, Eq, Ord)
+
+data EcosystemSpecific = EcosystemSpecific
+  { ecosystemSpecificAffects :: Affects
+  } deriving (Show, Eq, Ord)
+
+data Events = Events
+  { eventsIntroduced :: Text
```
mihaimaruseac, line 36: Should we also include the other 3 types of events? Even if we don't use them right now?

All of these would be optional.

https://ossf.github.io/osv-schema/#affectedrangesevents-fields

code/hsec-tools/src/Security/OSV.hs
mihaimaruseac: +1

code/hsec-tools/src/Security/OSV.hs
mihaimaruseac: I think we should include `versions`, there are packages not using SemVer in Hackage and [recommendation](https://ossf.github.io/osv-schema/#affectedrangestype-field) is to populate the `versions` field whenever `ranges.type` is not "SEMVER".

code/hsec-tools/src/Security/OSV.hs
mihaimaruseac: +1

code/hsec-tools/src/Security/OSV.hs
mihaimaruseac: Also, this seems to miss `severity` field

https://ossf.github.io/osv-schema/#affectedseverity-field

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+  } deriving (Show, Eq, Ord)
+
+data Model = Model
+  { modelDetails :: Text
+  , modelId :: Text
+  , modelSummary :: Text
+  , modelRelated :: [Value]
```
mihaimaruseac, line 43: This is a list of `Text`?

https://ossf.github.io/osv-schema/#related-field

code/hsec-tools/src/Security/OSV.hs
```diff
@@ -0,0 +1,223 @@
...
+  , modelAffected :: [Affected]
+  , modelAliases :: [Value]
```
mihaimaruseac, line 45: Same as `modelRelated`, should be list of `Text`?

code/hsec-tools/src/Security/OSV.hs
mihaimaruseac: Also missing `withdrawn` and a general `database_specific` field (that applies to the entire vuln, not to a specific affected version).

https://ossf.github.io/osv-schema/#withdrawn-field
https://ossf.github.io/osv-schema/#database_specific-field

mihaimaruseac (commented):
Thanks for doing this! I've been slacking off on this as I was OOOish

frasertweedale:
@TristanCacqueray I'm not sure if you had work in progress, but I've created a new PR (#42) based on yours, and addressed several of the issues mentioned in review (but not all of them, yet).

code/hsec-tools/src/Security/OSV.hs
frasertweedale: I added these in #42 (which is based on this PR).

code/hsec-tools/src/Security/OSV.hs
frasertweedale: I changed it in #42.

code/hsec-tools/src/Security/OSV.hs
frasertweedale: I changed it in #42.

code/hsec-tools/src/Security/OSV.hs
frasertweedale: I addressed this in #42.

frasertweedale (commented):
Added some comments about what has been addressed (so far) in #42.

frasertweedale:
Closing (#42, which extended this PR, was merged).

PR #38, opened 2023-06-13
ci: check HSEC id uniqueness

frasertweedale:
Add a CI check that HSEC ids (or the filenames, at least) are unique.

I pushed a branch with a conflict to test the new CI step.  Result:
https://github.com/frasertweedale/security-advisories/actions/runs/5251623664/jobs/9486797411


blackheaven:
I think we should also check toml's `id` too

frasertweedale:
> I think we should also check toml's `id` too

I plan on adding this to the `hsec-tools check` command (will change it to read a named file, instead of `stdin`).  So that will be done in a separate PR, when I (or someone else) get to it.

blackheaven (approved):

PR #39, opened 2023-06-13
Parse summary from markdown

divanorama:
---

## Advisory

- [ ] It's not duplicated
- [ ] All fields are filled

## hsec-tools


TristanCacqueray (approved):

frasertweedale:
It makes good sense.  Thanks for contributing!

Issue #43, opened 2023-06-15
tools: switch to pandoc AST


frasertweedale:
Per SRT discussions, we will switch away from the handrolled document AST to the *pandoc-types* AST.  [*commonmark-pandoc*](https://hackage.haskell.org/package/commonmark-pandoc) provides the *commonmark* integration.

PR #44, opened 2023-06-17
lib/advisories: parse into pandoc AST

frasertweedale:
Instead of our hand-rolled document AST, parse the advisory content into a `Pandoc` document via *commonmark-pandoc*.  Store the `Pandoc` value in the `advisoryPandoc` field.

We continue (for now) to retain the `advisoryHtml` field.  It now displays the raw TOML block instead of a prettier HTML table - a mild regression.  But we expect to remove this field in the future, when we implement full site rendering for the advisory database.

Fixes: https://github.com/haskell/security-advisories/issues/43

### DIFF
--- a/code/hsec-tools/hsec-tools.cabal
+++ b/code/hsec-tools/hsec-tools.cabal
@@ -38,7 +38,9 @@ library
                       containers >= 0.6 && < 0.7,
                       commonmark ^>= 0.2.2,
                       toml-reader ^>= 0.1 || ^>= 0.2,
-                      aeson >= 2
+                      aeson >= 2,
+                      pandoc-types >= 1.22 && < 2,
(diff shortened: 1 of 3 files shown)


code/hsec-tools/hsec-tools.cabal
```diff
@@ -38,7 +38,9 @@ library
                       containers >= 0.6 && < 0.7,
                       commonmark ^>= 0.2.2,
                       toml-reader ^>= 0.1 || ^>= 0.2,
-                      aeson >= 2
+                      aeson >= 2,
+                      pandoc-types >= 1.23 && < 2,
```
blackheaven, line 42:
```suggestion
                      pandoc-types >= 1.22 && < 2,
```

for nix

blackheaven (approved):

blackheaven (approved):

Issue #32, opened 2023-06-10
Add initial content


TristanCacqueray:
The goal is to add some real advisory for known or past issue.

- [ ] Follow the documentation and propose new advisory.
- [ ] Update documentation if necessary.

frasertweedale:
Search of NVD turned up the following CVEs:

- https://nvd.nist.gov/vuln/detail/CVE-2022-31053 biscuit-haskell improper sig validation
- https://nvd.nist.gov/vuln/detail/CVE-2013-1436 xmonad-contrib code injection
- https://nvd.nist.gov/vuln/detail/CVE-2021-30502 vscode-ghc-simple RCE
- https://nvd.nist.gov/vuln/detail/CVE-2021-4249 xml-conduit entity expansion DoS
- https://nvd.nist.gov/vuln/detail/CVE-2013-0243 hs-tls improper certificate validation
- https://nvd.nist.gov/vuln/detail/CVE-2022-3433 aeson hash flooding (#35)

frasertweedale:
@david-christiansen  has a known TOML lib issue (already fixed) that he will submit next week.

frasertweedale:
Re https://nvd.nist.gov/vuln/detail/CVE-2021-30502 vscode-ghc-simple RCE - it is actually not a Haskell program.  I wonder if it has an advisory published in some other ecosystem, other than the "bare" CVE?

mihaimaruseac:
+1, I don't think we have to do anything more for this one

frasertweedale:
I think we're done here :)  All the known historical advisories have been added to the DB.

PR #47, opened 2023-06-17
Report biscuit-haskell CVE-2022-31053 (#32)

blackheaven:
---

## Advisory

- [x] It's not duplicated
- [x] All fields are filled

### DIFF
--- a/advisories/hackage/biscuit-haskell/HSEC-2022-0001.md
+++ b/advisories/hackage/biscuit-haskell/HSEC-2022-0001.md
@@ -1,0 +1,29 @@
+```toml
+[advisory]
+id = "HSEC-2022-0001"
+package = "biscuit-haskell"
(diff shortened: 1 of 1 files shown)


mihaimaruseac (approved):

frasertweedale:
Need to bring something up...

I thought we had decided that HSEC IDs would be assigned based on year the advisory was created - not the year in which the problem the advisory describes became known.  Therefore, shouldn't this have been **HSEC-2023-0002**?

We have not yet exported any OSV objects or other enduring artifacts from our DB.  So if the SRT agrees that we should rename it, it would be safe to do so.  Let me know what you think.

mihaimaruseac:
Oh, I forgot about that. Yes, I'm ok with switching to 2023.

blackheaven:
Fixed in #55

PR #54, opened 2023-06-19
Report xmonad-contrib CVE-2013-1436

akacase:
---

## Advisory

- [x] It's not duplicated
- [x] All fields are filled
- [x] It is validated by `hsec-tools`

PR #55, opened 2023-06-19
fix(advisory): rename HSEC-2022-0001 to HSEC-2023-0002

blackheaven:
---

## Advisory

- [x] It's not duplicated
- [x] All fields are filled


akacase:
@blackheaven by renaming this file to `0002` this conflicts with #54. Do you want me to rename to `0003`?

frasertweedale:
Thanks @blackheaven !

Issue #53, opened 2023-06-19
No specification of "references" section


david-christiansen:
## Summary

In the advisory template, there's a `references` section in the front matter:
```
[[references]]
type = "ARTICLE"
url = "https://example.com"
```

However, neither this section nor the extended example with all the comments describes what the valid entries for the `type` field are.

frasertweedale:
@david-christiansen the README has some explanation.  Feel free to copy/paste it to other relevant places.
Or someone else will get to it soon :)

blackheaven:
I'll add an explanation in the cli error

david-christiansen:
This comes from me role-playing a contributor, and they'll not be doing the command-line, right? The expected workflow is a pull request.

I'm a bit skeptical of copy-pasting information around, rather than having a single canonical source - is this perhaps evidence that the format docs are spread too much around and could be reorganized into a single file?

blackheaven:
I mean, I am serious, I came across this pitfall doing mine, but I was too lazy to fix it (to read fully the documentation too).

I agree you can't ask the contributor to run the check locally, but accurate/contextualized error messages is a way to ease contribution and would be displayed in the CI run result.

david-christiansen:
I just got it in the CI, and I now understand what you meant. Good point!

I think that it would also be useful for the CI to not show all the ones that pass, or to at least split passing and failing files in the output. I imagine that when we have 100 entries, it'll be hard to find the error messages to act on.

david-christiansen:
(sorry if the smiley came across as sarcastic, I was trying to communicate "yay, someone's doing it" not "what a silly answer")

blackheaven:
I may have been a bit defensive, sorry for that (I thought you didn't thought it was that of a big deal since you discovered during role-playing).

frasertweedale:
For sure, it would be good to have a bot that can give richer feedback in PRs to help contributors.  For now, we should just improve the docs, because there are still quite a few tooling enhancements on the critical path (OSV export, website builder, etc).

david-christiansen:
I think that this improvement to the CI is great, as most people don't proactively read docs. This will help. Thanks @blackheaven!

PR #56, opened 2023-06-19
Add an advisory for a potential DOS in toml-reader-0.1.0.0

david-christiansen:
This is my attempt to validate the process by creating an advisory as if I were not associated with the project. I've opened a couple of PRs and issues based on what I found. The rest of this PR test will be me role-playing through the process.

-------------------------

This is an advisory for an issue from `toml-reader` that I found some time ago. I [reported it to the author of the package](https://github.com/brandonchinn178/toml-reader/issues/8) and he promptly [fixed it](https://github.com/brandonchinn178/toml-reader/pull/9). I emailed him about creating this advisory, and he's aware that I'm doing so and supports it.

See the advisory text for details, but the basic problem is that the parser for numbers takes time linear in the size of the denoted number. Because of exponent notation, this means that a string of length _n_ can take **O**(10^_n_) time to parse.

I don't really understand CVSS, so I left it blank, hoping that the committee will help me create it. From what I can see, this is only a vulnerability if the library is used to parse untrusted input. If used for the common case, which is a configuration file, then the problem is no big deal, as someone being able to DOS their own machine doesn't let them hurt others. But the very tools that underlie this database use the library to parse TOML that's coming from others, so a DOS is indeed a concern.

---

## Advisory

- [x] It's not duplicated
- [x] All fields are filled

## hsec-tools


david-christiansen:
I'm not sure what I'm doing wrong - I got the following validation error:
```
advisories/hackage/biscuit-haskell/HSEC-2022-0001.md: no error
Advisory structure error:
InvalidFormat "reference.type" "BUG REPORT"
advisories/hackage/toml-reader/HSEC-0000-0000.md: advisories/hackage/aeson/HSEC-2023-0001.md: no error
EXAMPLE_README.md: no error
EXAMPLE_ADVISORY.md: no error
```
The message seems to be telling me about lots of files that I didn't just create, plus the one that I did. I guess it means that the others are fine, but the one I created has a bad reference type?

david-christiansen:
Someone in an issue (#53) pointed out that the example advisory in the README has the allowed categories. I'll revise!

david-christiansen:
Meta comment: it might be worth making a little command-line app that gives a somewhat friendlier way to create these advisories. I think that most Haskell users don't really know what a CWE or CVSS is, and we'll most commonly get fairly minimal TOML blocks rather than people going out and reading up. One way to address this could be a little script in the repo that asks a few questions and guides them through this process.

david-christiansen:
My CI has now passed, but it looks to me like the job that validates the advisory didn't actually run! In particular, I don't see "Invoke check-advisories workflow" as one of the jobs behind the green check on [6cfeccc](https://github.com/haskell/security-advisories/pull/56/commits/6cfeccc54c99e3b4c491f455944171b2df293d0f), but it's what failed on [6299888](https://github.com/haskell/security-advisories/pull/56/commits/6299888375844454c4b3c0c89e8652eb4701c385).

frasertweedale:
> My CI has now passed, but it looks to me like the job that validates the advisory didn't actually run! In particular, I don't see "Invoke check-advisories workflow" as one of the jobs behind the green check on [6cfeccc](https://github.com/haskell/security-advisories/pull/56/commits/6cfeccc54c99e3b4c491f455944171b2df293d0f), but it's what failed on [6299888](https://github.com/haskell/security-advisories/pull/56/commits/6299888375844454c4b3c0c89e8652eb4701c385).

It did run on [6cfeccc](https://github.com/haskell/security-advisories/pull/56/commits/6cfeccc54c99e3b4c491f455944171b2df293d0f): https://github.com/haskell/security-advisories/actions/runs/5315187694/jobs/9623351121

If I'm not mistaken, the `hsec-tools check` job does not appear until after the Haskell-CI matrix jobs have completed (successfully).

david-christiansen:
Ah, now I see it. I suppose I was just insufficiently patient (I hit reload a few times right after they stopped running and it had turned green and didn't see it). This was also in the evening, and coffee had not been consumed for some time. In any case, thanks!

david-christiansen:
Is there anything else I need to do with this report?

TristanCacqueray (approved):

blackheaven (approved):
Actually you need to rename it `HSEC-2023-0004.md`

BTW, it's something the bot should do.

TristanCacqueray:
Renaming is annoying, the `CONTRIBUTING.md` doesn't explain how to work exactly. Perhaps we could adapt https://github.com/rustsec/advisory-db/blob/main/.github/workflows/assign-ids.yml ?

frasertweedale:
Yeah, I think we could do something *similar* to the rust job.  One thing I do object to in the RustSec approach is that advisories get merged to main with ID `RUSTSEC-0000-0000`, *then* the job kicks in and creates a *second* PR to assign the ID.  I would prefer an approach that checks and updates the first initial pull request, so that everything that lands on `main` has a valid ID.  (This is important for the OSV export job, so that we do not generate and export bad data.)

david-christiansen:
It looks to me like [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) could solve this - it looks like a kind of integrated marge-bot or homu.

Perhaps a simple solution would be an action that assigns an ID  (perhaps built by composing https://github.com/marketplace/actions/file-changes-action with a tiny blob of shell to rename new files named `HSEC-0000-0000.md` to the right thing), triggered by a PR being added to a merge queue?

david-christiansen:
I renamed the file. I did it in two steps to see if the CI catches the mismatch between the filename and the internal ID :)

david-christiansen:
Back to my user role-play: How should I interpret this?
```
advisories/hackage/biscuit-haskell/HSEC-2023-0002.md: no error
advisories/hackage/xmonad-contrib/HSEC-2023-0003.md: no error
advisories/hackage/aeson/HSEC-2023-0001.md: no error
Advisory structure error:
IllegalOutOfBandOverride "date"
advisories/hackage/toml-reader/HSEC-2023-0004.md: EXAMPLE_README.md: no error
EXAMPLE_ADVISORY.md: no error
```

frasertweedale:
> Back to my user role-play: How should I interpret this?
> 
> ```
> advisories/hackage/biscuit-haskell/HSEC-2023-0002.md: no error
> advisories/hackage/xmonad-contrib/HSEC-2023-0003.md: no error
> advisories/hackage/aeson/HSEC-2023-0001.md: no error
> Advisory structure error:
> IllegalOutOfBandOverride "date"
> advisories/hackage/toml-reader/HSEC-2023-0004.md: EXAMPLE_README.md: no error
> EXAMPLE_ADVISORY.md: no error
> ```

You have to remove the `date` field from the TOML.  That date of the advisory will be inferred from the Git history.

frasertweedale:
> Edit: Shell is not enough - we'll probably need some Haskell to reassign the ID in the TOML block. I suppose the command-line tools could get an argument that makes them assign missing IDs in the repo.

The `rustsec-admin` tool has a command to assign IDs to unassigned advisories.  We can do something similar.  It involves both editing the TOML and renaming files.  Most likely I will just search-and-replace to edit the TOML rather than using the first-class TOML serialisation, to avoid field re-ordering (or the complexity we could incur to avoid it).

frasertweedale:
I'll take over this PR.  It will be the first advisory to include multiple affected libraries (*base*, *toml-reader*, ...?).  We have some *hsec-tools* work to do to support that scenario.  I propose to use symbolic links to reference it from multiple directories (with the hard copy in *base*, probably).

See also the discussion at https://gitlab.haskell.org/ghc/ghc/-/issues/23538

frasertweedale:
I updated the PR - it now mentions both *base* and *toml-reader* as affected.  It has also been assigned id **HSEC-2023-0007**

frasertweedale:
Failed uniqueness check is expected, and will require rebase after #98 lands.

mihaimaruseac (approved):

PR #66, opened 2023-06-23
deps: Cabal -> Cabal-syntax

frasertweedale:
We can narrow our dependency from *Cabal* to *Cabal-syntax*.


blackheaven (approved):

frasertweedale:
Nix doesn't like it due to incompatibility with (the version of cabal pinned to) GHC 9.2.  I'll drop this for now.

blackheaven:
It's `3.6.0.0` on nixpkgs, it should work, otherwise, we can override the flake.

frasertweedale:
> It's `3.6.0.0` on nixpkgs, it should work, otherwise, we can override the flake.

I don't want to spend more time on it right now.  If you want to have a crack at it, go for it :)

PR #68, opened 2023-06-23
deps: Cabal -> Cabal-syntax

blackheaven:
Reboot of #66

---

## hsec-tools


frasertweedale:
Thanks @blackheaven!

PR #69, opened 2023-06-24
fix(tool): rely on nixos-unstable to use cache

blackheaven:
---

## hsec-tools

### DIFF
--- a/code/hsec-tools/flake.lock
+++ b/code/hsec-tools/flake.lock
@@ -20,15 +20,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1687289913,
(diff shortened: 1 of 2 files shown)


blackheaven:
I wasn't ready :/

nix takes way to long to build, there is a cache issue I'm trying to fix.

akacase:
@blackheaven might want to try a rev from last week on the pin, as it takes a while for all the Haskell packages to be built and populated; it'll keep us passing the flake-check as well.

blackheaven:
Actually, I thought that `nixos-unstable` had all it built, I 'll have a look because on my machine I had cache, but I have additional caches.

PR #71, opened 2023-06-26
fix(tool): speedup nix build time WIP

blackheaven:
---

## hsec-tools

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -13,6 +13,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Install Nix
         uses: DeterminateSystems/nix-installer-action@main
+      - name: Run the Magic Nix Cache
+        uses: DeterminateSystems/magic-nix-cache-action@v1
       - name: Check Nix flake inputs
         uses: DeterminateSystems/flake-checker-action@v4
         with:
--- a/code/hsec-tools/.envrc
+++ b/code/hsec-tools/.envrc
@@ -1,4 +1,4 @@
-use flake
+use flake --accept-flake-config
 
 if [ -f .env ]; then
   . .env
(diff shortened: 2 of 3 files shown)


frasertweedale:
I'll leave this for the nix folks to review, lest I merge it too early like last time :]

akacase:
@blackheaven https://determinate.systems/posts/magic-nix-cache -- :)

blackheaven:
@akacase Thanks, I hope it won't be too big.

However I think we need to use a cachix, @david-christiansen does HF provide one for haskell projects?

david-christiansen:
The HF does not provide that service. What is it? What is the value? What would it cost?

TristanCacqueray:
@david-christiansen it's a service to cache ci nix build (e.g. the build-depends), and the value is to greatly improve ci run time. There is a free tier for OSS projects, but the quota is bound to a single user or org.

@blackheaven perhaps we could try (announced today) https://determinate.systems/posts/magic-nix-cache ?

blackheaven:
@TristanCacqueray @akacase just suggested it

I just tested it, we go from 42m to 30m, not a real improvement IMO.

I think the only viable option would be to setup cachix, but I don't feel like pushing for it as it's not currently setup and I guess agreeing on the tooling at HF-level is out of the scope of this PR.

david-christiansen:
If it's a matter of paying an affordable monthly subscription, then HF can definitely provide it. We don't need to agree to use something for every project - it's fine if it's useful for one.

I don't want to be administering any servers, though - better to pay a bit of money and have something that just works.

blackheaven:
Sure, but I don't think we are making an intense use of nix (for the moment :crossed_fingers:), let's reevaluate this in few weeks.

PR #73, opened 2023-06-27
implement OSV conversion

frasertweedale:
This PR:

- reverts the "limit" change (pursuant to discussion in https://github.com/haskell/security-advisories/issues/3#issuecomment-1609061783)
- improves the documentation of the `affected.versions` table
- implements OSV conversion in the library
- add the `hsec-tools osv` subcommand for generating OSV data

```
commit 067ac75ea7a495cdb0c5e2fa9f0f38fa58f876eb
Author: Fraser Tweedale <frase@frase.id.au>
Date:   Tue Jun 27 22:00:21 2023 +1000

    Revert "lib/advisories: add "limit" field for version ranges"
    
    This reverts commit 6b95156a5f373f0cf87576b1ce155c3adc92d479.
    
    Per discussion at [1], the OSV project prefers to (ab?)use the
    "fixed" event type to limit a range, even when there is no fix.
    Revert the commit.  We will add better documentation in a subsequent
    commit.
    
    [1]: https://github.com/haskell/security-advisories/issues/3#issuecomment-1609066385

commit 1c0fd46ce2cbc262e9d9f895b6d52b1bb2230c29
Author: Fraser Tweedale <frase@frase.id.au>
Date:   Tue Jun 27 22:14:40 2023 +1000

    docs: more explanation of affected.versions

commit 2121be02c5bd8757892d9e160f46535f5b9e0373
Author: Fraser Tweedale <frase@frase.id.au>
Date:   Tue Jun 27 22:31:39 2023 +1000

    tools: implement OSV conversion
    
    Add the `Security.Advisories.Convert.OSV` module, which defines the
    conversion from our `Advisory` data type to the OSV `Model`.
    Currently, no database-specific or ecosystem-specific fields are
    set.  Whether or how to use those fields is a matter for future
    discussion.
    
    Re-export the *aeson* encode and decode functions from
    `Security.OSV`, for convenience.
    
    Add the `osv` subcommand to `hsec-tools`.  It works in the same way
    as `check`, but emits the encoded OSV JSON data.
    
    Later commits will add the CI workflows to generate and publish the
    OSV data.
    
    Fixes: https://github.com/haskell/security-advisories/issues/3
```


code/hsec-tools/src/Security/OSV.hs
```diff
@@ -24,6 +24,11 @@ module Security.OSV
   , ReferenceType(..)
   , referenceTypes
   , Severity(..)
+
+  -- * /aeson/ re-exports
+  , encode
+  , eitherDecode
+  , eitherDecode'
```
blackheaven, line 31: why do you reexport them?

frasertweedale: Just for convenience... if you're using `Security.OSV`, it's quite likely you'll be decoding from, or encoding to JSON.  Avoids an explicit import - and perhaps for some use cases it could avoid a direct dependency on *aeson*.

blackheaven: I find it quite confusing, maybe it'll be clearer when it'll be in a library on its own.

code/hsec-tools/app/Main.hs
```diff
@@ -44,6 +47,16 @@ commandCheck =
...
       T.putStrLn "no error"
 
+commandOsv :: Parser (IO ())
+commandOsv =
+  withAdvisory go
+  <$> optional (argument str (metavar "FILE"))
+  <**> helper
+  where
+    go _ adv = do
+      L.putStr (OSV.encode (OSV.convert adv))
```
frasertweedale, line 57: @blackheaven this is where the re-export pays off.  I didn't have to explicitly import `Data.Aeson.encode` to use it here.

If you still prefer to avoid the re-export and instead import `Data.Aeson` directly, I'm OK with that.  I can make the change and update the PR tomorrow.

TristanCacqueray (approved):

mihaimaruseac (approved):

PR #77, opened 2023-06-28
Add PROCESS.md

TristanCacqueray:
This PR imports the OpenStack VMT's process. The main differences are:
- We don't use an issue tracker with an easy to follow status fields or tags.
- Our mail addresses doesn't have associated gpg keys.

Thus I removed these parts from the original vmt-process, but perhaps we could also adopt these practices.

Moreover, we don't have a downstream stakeholder list yet. At the last meeting we talked about adding the relevant mails addresses to the repository, but we might want to consider a `embargo-notice@haskell.org` list instead.

Lastly, the vmt-process defines an `impact description` that is a subset of the final advisory. It seems like we could do the same by omitting the CVE and references. That way we can refer to the document that is being created before the CVE request.

### DIFF
--- a/PROCESS.md
+++ b/PROCESS.md
@@ -6,8 +6,9 @@
 >   http://creativecommons.org/licenses/by/3.0/legalcode
 > Original version: https://opendev.org/openstack/ossa/src/branch/master/doc/source/vmt-process.rst
 
-The Haskell Security Response Team (HSRT) is responsible for coordinating
-the progressive disclosure of a vulnerability.
+The HSRT coordinates responsible disclosure of security vulnerabilities in
+the Haskell ecosystem and help Haskell users have an accurate and timely
+assessment of risks.
 
 Members of the team are independent and security-minded folks who ensure
 that vulnerabilities are dealt with in a timely manner and that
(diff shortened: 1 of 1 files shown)


PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+mitigation mechanisms. The HSRT coordinator should use the template
+below.
+
+### Review advisory
+
+The advisory is validated by the reporter and the project's maintainer.
+
+### Send CVE request
+
+To ensure full traceability, we attempt to obtain a CVE assignment
```
frasertweedale, line 89: Do we?  Should we?

We assign our own IDs.  Of course, we record corresponding CVE IDs (if any exist) in the `aliases` field.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+
+MITRE returns the assigned CVE. It is added to the advisory,
+and the bug is retitled to `$TITLE ($CVE)`.
+
+### Embargoed disclosure
+
+Once the patches are approved and the CVE is assigned, a signed email
+with the vulnerability description is sent to the downstream
+stakeholders. The disclosure date is set to 3-5 business days, excluding
+Monday/Friday and holiday periods, at 1500 UTC. No stakeholder is
```
frasertweedale, line 131: This seems too specific for us, at this stage.  Maybe we can just say that the notice will state the planned disclosure date.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+
+Shortly after pushing the patches (potentially waiting for the first
+test runs to complete), publish the advisory to the repository.
+
+### All patches merged
+
+Patches approved in code review do not necessarily merge immediately,
+but should be tracked closely until they do (if the bug number is
+correctly identified in commit messages then it will be automatically
+updated to reflect this as well). Subsequent security point releases of
```
frasertweedale, line 168: This seems OpenShift specific; I do not think we will have any such automation.

frasertweedale (commented):
Some initial comments.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+downstream stakeholders are notified in a coordinated and fair manner.
+Where a member of the team is employed by a downstream stakeholder, the
+member does not give their employer prior notice of any vulnerabilities.
+In order to reduce the disclosure of vulnerability in the early stages,
+membership of this team is intentionally limited to a small number of
+people.
+
+## Supported versions
+
+The Vulnerability Management team coordinates patches fixing
```
david-christiansen, line 23:
```suggestion
The SRT coordinates patches fixing
```

david-christiansen, line 23: I actually think that we should say that the role of the team is to coordinate responsible disclosure of security vulnerabilities in the Haskell ecosystem and help Haskell users have an accurate and timely assessment of risks.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+vulnerabilities in packages published via Hackage.
+
+## Process
+
+Each security bug is assigned a HSRT *coordinator* (member from the
+haskell security response team) that will drive the fixing and
```
david-christiansen, line 29:
```suggestion
Haskell Security Response Team) that will coordinate the
```
We don't fix the bugs, we just help inform people

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+-   check that the report indicates the correct project and adjust as
+    necessary.
+-   contact the project's maintainer for confirmation
+    of impact and determination of affected version.
+
+### Patch development
+
+For a private report, the reporter (automatic if reported directly as a
+bug) and the affected projects' maintainers plus anyone they deem necessary
+to develop and validate a fix are added to the bug's notified list.
```
david-christiansen, line 51: What is a notified list?

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+**not sent to the public code review system**.
+
+For public reports, there is no need to directly subscribe anyone and
+patches can be submitted directly to the code review system instead of
+as bug attachments (though the bug should be referenced in any commit
+messages so it will be updated automatically).
+
+If project-side delays are encountered at this or any subsequent stage
+of the process, the HSRT and other interested parties may reach out to
+that project's maintainers requesting more immediate attention to the issue.
```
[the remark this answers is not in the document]
david-christiansen: I agree - I 'd replace this section.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Patch review
```
david-christiansen, line 64: I'd get rid of this section - we're not going to be reviewing everyone's patches

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+For a private report once the initial patch has been attached to the
+bug, core reviewers on the subscription list from the project in
+question should review it and suggest updates or pre-approve it for
+merging. Privately-developed patches need to be pre-approved so that
+they can be fast-tracked through public code review later at disclosure
+time.
+
+### Draft advisory
+
+In the mean time, the HSRT coordinator prepares a security advisory
```
david-christiansen, line 75: I think that we should work with the person who reports it to refine their draft security advisory, rather than writing them ourselves.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Open bug, Push patch
+
+In preparation for this, make sure you have a maintainer available to
+help pushing the fix at disclosure time.
+
+On the disclosure hour, open bug, push patches for review and
+fast-track approvals (referencing the bug).
+
+Update the bug title to `[HSEC-$NUM] $TITLE`.
```
david-christiansen, line 147: I think this isn't relevant, as we don't have the bug trackers.

PROCESS.md
```diff
@@ -0,0 +1,279 @@
...
+Where a member of the team is employed by a downstream stakeholder, the
+member does not give their employer prior notice of any vulnerabilities.
+In order to reduce the disclosure of vulnerability in the early stages,
+membership of this team is intentionally limited to a small number of
+people.
+
+## Supported versions
+
+The Vulnerability Management team coordinates patches fixing
+vulnerabilities in packages published via Hackage.
```
blackheaven, line 24: I wonder if it's not too restrictive.

PROCESS.md
TristanCacqueray: That would be the issue notification/participant list, but I'm removing this part since it does not apply for us.

PROCESS.md
TristanCacqueray: Following what rustsec does, we decided that CVE assignment is optional. It's already documented in the CONTRIBUTING doc, so I removed that part.

TristanCacqueray:
Thank you for the feedback. I've revised the document accordingly in 005d8a9 , and I've removed the embargo related bits in fd46bd4 since we decided to not support this process for now.

I think it would be nice to offer such service, but the OpenStack process would need to be further adapted because our scope is bigger. For example, Haskell packages does not share a common issue tracker we could use to coordinate the patch development. For example, in the advance notice we don't really have a way to subscribe downstream stakeholder to the issue because it is presently a mailing-list thread. Therefore, we might want to remove the last section, and perhaps merge this document in the CONTRIBUTING.md.

That being said, I also think it is a good starting point.

blackheaven (approved):

frasertweedale:
> That being said, I also think it is a good starting point.

I agree.  There will be some refinements but the major things identified in the first pass have been resolved.  So let's merge it and refine it over time.

PR #80, opened 2023-06-29
Improve pipelines

blackheaven:
- [x] Save latest `hsec-tools` as artifact
- [x] Reuse artifact to check new advisories
  - [x] Only run on new/changed advisories
- [x] Don't rebuild `hsec-tools` on new advisories

For reference, I squeezed the docker image: https://gautier.difolco.dev/2023-07/nix-docker-optimization/


mihaimaruseac (approved):

.github/workflows/haskell-ci.yml
```diff
@@ -17,8 +17,23 @@ on:
...
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          concurrent_skipping: "never"
+          skip_after_successful_duplicate: "true"
+          paths: '["code/**"]'
+          do_not_skip: '["push", "workflow_dispatch", "schedule"]'
   linux:
     name: Haskell-CI - Linux - ${{ matrix.compiler }}
+    needs: tools_changed
+    if: ${{ needs.tools_changed.outputs.should_skip != 'true' }}
```
frasertweedale, line 36: Whatever changes we make in `haskell-ci.yml`, they need to be implemented in `.github/haskell-ci.patch` so that `haskell-ci regenerate` will preserve the modifications/additions.  See https://frasertweedale.github.io/blog-fp/posts/2023-06-04-haskell-ci-use-executable.html#patching-haskell-ci.yml for details.

.github/workflows/haskell-ci.yml
```diff
@@ -17,8 +17,23 @@ on:
...
   - pull_request
 jobs:
+  tools_changed:
+    continue-on-error: true
+    runs-on: ubuntu-20.04
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
```
frasertweedale, line 27: Should we pin a particular release tag, instead of `master`?

.github/workflows/check-advisories-standalone.yml
```diff
@@ -0,0 +1,38 @@
...
+          skip_after_successful_duplicate: "true"
+          paths: '["advisories/**"]'
+          do_not_skip: '["push", "workflow_dispatch", "schedule"]'
+  check-advisories:
+    name: Invoke check-advisories workflow
+    needs: [tools_changed, advisories_changed]
+    if: ${{ needs.tools_changed.outputs.should_skip == 'true' && needs.advisories_changed.outputs.should_skip != 'true' }}
+    uses: ./.github/workflows/check-advisories.yml
+    with:
+      artifact-name: hsec-tools-main
```
frasertweedale, line 37: This artifact gets uploaded by the nix job, which runs concurrently.  As a result, this job probably runs *before* the nix job completes.  As a consequent, it might be using an out-of-date version of hsec-tools, which could result in incorrect CI results.

.github/workflows/haskell-ci.yml
frasertweedale, line 36: Note that the expected artifact name for this workflow is:
```yaml
with:
      artifact-name: hsec-tools-${{ github.sha }}
```

If we skip the `linux` job when `code/**` did not change, it would fail to upload the artifact, and the `check-advisories` workflow would fail.  To avoid this, we can base the artifact name on the Git tree hash for the `code/` directory, instead of the commit hash.

I'll update this commit with the Git command to query the tree hash (I wrote it down somewhere, just need to find it).

mihaimaruseac (approved):

.github/workflows/check-advisories-standalone.yml
blackheaven: I'm not sure it can happen, I used the 'skip' to make it exclusive with the nix  job.

.github/workflows/haskell-ci.yml
blackheaven: IIUC, it's only used later-on to check advisories in this workflow, I'm not sure it's used elsewhere

blackheaven:
Eventually!

@frasertweedale can you have a look please?

akacase:
@blackheaven -- This is great! I like the docker addition. Should we strive for parity by also adding a Dockerfile in the main repository to replicate (sort of) what we have in Nix?

blackheaven:
It could be a good idea.

At some point, we also might publish GitHub packages too.

TristanCacqueray (approved):

frasertweedale:
@blackheaven I promise I haven't forgotten about this PR :)  I shall re-review it this week.

mihaimaruseac (approved):

mihaimaruseac:
We need to resolve the nix conflict.

blackheaven:
I'll handle it tomorrow morning

blackheaven:
Ready :)

akacase (approved):

frasertweedale:
`haskell-ci regenerate` fails because the patch does not apply.  (I'm using haskell-ci v1.6.3).  I'll look into it tomorrow, if I can find some time.

frasertweedale (changes requested):
I *think* the following `haskell-ci.patch` will achieve the desired result (that is, regenerates and cleanly applies the patch to `haskell-ci.yml` resulting unchanged file content).

```patch
diff --git a/.github/workflows/haskell-ci.yml b/.github/workflows/haskell-ci.yml
index 96282c2..e40c573 100644
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -19,6 +19,21 @@
 jobs:
+  tools_changed:
+    continue-on-error: true
+    runs-on: ubuntu-20.04
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@v5.3.0
+        with:
+          concurrent_skipping: "never"
+          skip_after_successful_duplicate: "true"
+          paths: '["code/**"]'
+          do_not_skip: '["push", "workflow_dispatch", "schedule"]'
   linux:
     name: Haskell-CI - Linux - ${{ matrix.compiler }}
+    needs: tools_changed
+    if: ${{ needs.tools_changed.outputs.should_skip != 'true' }}
     runs-on: ubuntu-20.04
     timeout-minutes:
       60
```

When the `regenerate` issue is resolved feel free to merge it.  I am still not 100% convinced that everything will work as intended (same doubts as in previous comments).  But it's hard to fit all the cases in my head, so I think it's best to merge and then we can observe the changes in action.

frasertweedale (approved):

blackheaven:
I'll handle `main` build failure

PR #94, opened 2023-07-19
HSEC-2023-0006: x509-validation ignores pathLenConstraint

frasertweedale:
Historical issue, not previously reported.  I noticed it while investigating HSEC-2023-0005 (tls-extra ignores BasicConstraints)

## hsec-tools


blackheaven (approved):

frasertweedale:
Ah, a bug in our parser.  References field should be optional.  I'll make a PR for that.

frasertweedale:
see #95.

I added a FIX reference to this advisory, anyway :)

TristanCacqueray (approved):

blackheaven:
nix cache isn't efficient

TristanCacqueray:
@blackheaven It seems like nix rebuild a lot of thing because of the Cabal-syntax overrides, perhaps we should pick cabal-install and language-server from the un-modified set?

PR #96, opened 2023-07-20
fix(tools): override dependencies at package level to improve caching in nix

blackheaven:
See https://github.com/haskell/security-advisories/pull/94#issuecomment-1642304387

---

## hsec-tools

### DIFF
--- a/code/hsec-tools/flake.nix
+++ b/code/hsec-tools/flake.nix
@@ -17,9 +17,7 @@
 
           haskellPackages = pkgs.haskell.packages.ghc925.override
             {
-              overrides = hself: hsuper: {
-                Cabal-syntax = hsuper.Cabal-syntax_3_8_1_0;
-              };
+              overrides = hself: hsuper: { };
             };
         in
         rec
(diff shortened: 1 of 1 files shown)


frasertweedale:
@blackheaven nix build took ~40 minutes.  Do we expect subsequent runs to be faster?  Or said another way: how do we know if this change helps? :)

akacase:
@blackheaven -- In my [branch](https://github.com/akacase/security-advisories), I swapped to use `haskellPackages.developPackage` with the current overrides in place, and it utilizes the default cachix cache (on my end). Still, we default to `ghc928,` which is what `nixos-unstable` defaults to. We're currently using `ghc925` as that's the override? If I pin `developPackage` against any other version like `ghc945, ghc961, ghc962` etc. We'll end up having to use more custom caching options. If you got it handled, it's chill; just throwing out what's worked for me; on my branch, we're at 4m per build.

TristanCacqueray:
@blackheaven would you mind checking @akacase branch, it seems to work well, e.g. running `nix run github:akacase/security-advisories?dir=code/hsec-tools` is pretty fast. Perhaps we could move the flake to the root dir to avoid having to pass the `?dir=` argument.

blackheaven:
@akacase Thanks so much, it makes sense, I let you create a PR.

PR #104, opened 2023-07-24
Move the flake at the root of the repository

TristanCacqueray:
This change simplifies the nix usage so that we can run:

  nix run github:haskell/security-advisories


flake.nix
```diff
@@ -16,7 +16,7 @@
           pkgs.haskellPackages.developPackage {
             inherit returnShellEnv;
             name = "hsec-tools";
-            root = ./.;
+            root = "${self}/code/hsec-tools";
```
blackheaven, line 19: `./code/hsec-tools` does not work?

blackheaven (approved):

blackheaven (changes requested):
There's also `.envrc` and `shell.nix`

akacase:
@TristanCacqueray while you're changing those we can reduce our usage on the `shell.nix` compat piece by just using builtins now via: `(builtins.getFlake ("git+file://" + toString ./.)).devShells.${builtins.currentSystem}.default`

flake.nix
TristanCacqueray: That works, TIL :-)

TristanCacqueray:
Thanks for the quick review! @akacase your shell.nix didn't work (error: attribute 'devShells' missing), so I modified it a bit.

blackheaven (approved):

akacase (approved):

PR #105, opened 2023-07-24
tools: add golden tests to validate the parser

### LINKED ISSUE #100: add tests for advisory parsing
Before we merge #88, we need to add regression tests for advisory parsing.  Example-based tests will be fine, and we should test various success scenarios (in particular: presence/absence of optional fields) as well as common failure scenarios.

TristanCacqueray:
Fixes #100

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -219,6 +219,9 @@ jobs:
       - name: build
         run: |
           $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --write-ghc-environment-files=always
+      - name: tests
+        run: |
+          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --test-show-details=direct
       - name: hlint
         run: |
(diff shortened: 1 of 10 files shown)


blackheaven (approved):

frasertweedale:
Thanks for kicking this off Tristan.  I'd rather not use golden tests, because the "golden" output will likely change frequently, especially as we are still in early stages of tool development.  It will be a headache updating the tests.

Could we instead use example-based tests of inputs that are expected to parse or not parse, with some simple content checks for the success scenarios? (e.g., lists contain the expected values or are the expected length)

TristanCacqueray:
@frasertweedale I don't get how this causes a headache to maintain since this is done automatically when running with `--test-option=--accept`. And when comparing new outputs, we have a chance to spot unexpected diff. It sounds more tedious to create and maintain the simple content checks.

However success scenarios do produce a bit of data, so it might be more efficient to use direct checks. Perhaps we could use both strategies?

TristanCacqueray:
@frasertweedale can we land this now?

frasertweedale:
Yes.  Changes will come with the TOML lib change, but no reason not to land this now and just update the other MR.  Better to have the tests in place sooner :)

PR #106, opened 2023-07-25
fix(ci): adapt publish osv data (path)

blackheaven:
---
## hsec-tools


.github/workflows/check-advisories.yml
```diff
@@ -67,7 +67,7 @@ jobs:
             ID=$(basename "$FILE" .md)
             YEAR=$(echo "$ID" | cut -d - -f 2)
             mkdir -p $DATA_DIR/$YEAR
-            docker run --rm -v $PWD:/advisories haskell/hsec-tools:latest /bin/hsec-tools osv "$FILE" > $DATA_DIR/$YEAR/$ID.json
+            docker run --rm -v $PWD:/advisories haskell/hsec-tools:latest /bin/hsec-tools osv "advisories/$FILE" > $DATA_DIR/$YEAR/$ID.json
           done < <(find advisories -type f -name "*.md")
```
frasertweedale, line 71: And now I see the `$PWD:/advisories` which is a bit confusing...

Perhaps we should change the container volume mount point from `/advisories` to `/security-advisories` or even just `/repo`, to distinguish it from the `advisories` *subdirectory* and (hopefully?) make this code a bit easier to follow.

frasertweedale: I do like the `${{ github.workspace }}` change, but I was proposing to change the directory name on the *container* side :)

.github/workflows/check-advisories.yml
```diff
@@ -67,7 +67,7 @@ jobs:
             ID=$(basename "$FILE" .md)
             YEAR=$(echo "$ID" | cut -d - -f 2)
             mkdir -p $DATA_DIR/$YEAR
-            docker run --rm -v ${{ github.workspace }}:/advisories haskell/hsec-tools:latest /bin/hsec-tools osv "advisories/$FILE" > $DATA_DIR/$YEAR/$ID.json
+            docker run --rm -v ${{ github.workspace }}:/advisories haskell/hsec-tools:latest /bin/hsec-tools osv "repo/$FILE" > $DATA_DIR/$YEAR/$ID.json
```
frasertweedale, line 70: Don't we also need to change:

    -v ${{ github.workspace }}:/advisories

into:

    -v ${{ github.workspace }}:/repo

?

frasertweedale:
LGTM, thanks! :ship:

blackheaven:
:'( still broken

frasertweedale:
Fixed it finally.  The problem was that `${{ github.workspace }}` was the wrong directory.  We check out the repo at `${{ github.workspace }}/source`.  I went back to using $PWD.

Issue #110, opened 2023-07-27
HSEC ID reservation


frasertweedale:
As with CVEs, it is useful to be able to reserve an advisory.

- [ ] Define a data model for storing reservations in the repo.  Reservations should not indicate which package(s) are affected.
- [ ] Add an `hsec-tools` command to easily reserve an ID.
- [ ] Add a check to CI that the set of currently reserved IDs does not collide with defined advisories.

frasertweedale:
## Proposed data model (please provide feedback)

- a reservation is a file `advisories/reserved/HSEC-YYYY-NNNN.md`
  - as a result, uniqueness check is trivial.  Just extend the existing check to include the `reserved/` subdirectory.
- the file is empty (e.g. can be brought into existence via `touch`)
  - This prevents people including information that should not be public yet.
  - Alternatively, it could include minimal TOML content (e.g. just the "id").  But this is more effort to implement and verify.

mihaimaruseac:
I like the idea of an empty file for reservation. It is simple, easy to check, easy to implement

PR #114, opened 2023-08-09
hsec-tools: add reserve command

frasertweedale:
***note*** this PR consists of several commits.  I recommend to review
them separately, in chronological order, as they step by step introduce
the supporting library code and then the feature.

This PR adds the `reserve-id` command, which requests or assigns an HSEC
ID reservation.  It can also commit the change, ready for the
contributor to push and create a pull request.

Reservations are empty files under advisories/reserved.

CI enforces uniqueness.  When reservations get replaced with a full
advisory the reservation file must be removed in the same commit.

Reservation files SHOULD be empty, but nothing enforces that (yet).

Fixes: https://github.com/haskell/security-advisories/issues/110

---

## hsec-tools

### DIFF
--- a/code/hsec-tools/app/Command/Reserve.hs
+++ b/code/hsec-tools/app/Command/Reserve.hs
@@ -2,12 +2,17 @@
 
 module Command.Reserve where
 
-import Control.Monad (unless)
+import Control.Monad (unless, when)
 import Data.Maybe (fromMaybe)
 import System.Exit (die)
 import System.FilePath ((</>), (<.>))
 
-import Security.Advisories.Git (getRepoRoot)
+import Security.Advisories.Git
+  ( add
+  , commit
+  , explainGitError
+  , getRepoRoot
+  )
 import Security.Advisories.HsecId
   ( placeholder
   , printHsecId
@@ -30,8 +35,11 @@ data IdMode
   -- ^ Use the next available ID.  This option is more likely to
   -- result in conflicts when submitting advisories or reservations.
 
-runReserveCommand :: Maybe FilePath -> IdMode -> IO ()
-runReserveCommand mPath idMode = do
+data CommitFlag = Commit | DoNotCommit
+  deriving (Eq)
+
+runReserveCommand :: Maybe FilePath -> IdMode -> CommitFlag -> IO ()
+runReserveCommand mPath idMode commitFlag = do
   let
     path = fromMaybe "." mPath
   repoPath <- getRepoRoot path >>= \case
@@ -52,3 +60,12 @@ runReserveCommand mPath idMode = do
     fileName = printHsecId hsid <.> "md"
     filePath = advisoriesPath </> dirNameReserved </> fileName
   writeFile filePath ""  -- write empty file
+
+  when (commitFlag == Commit) $ do
+    let msg = printHsecId hsid <> ": reserve id"
+    add repoPath [filePath] >>= \case
+      Left e -> die $ "Failed to update Git index: " <> explainGitError e
+      Right _ -> pure ()
+    commit repoPath msg >>= \case
+      Left e -> die $ "Failed to create Git commit: " <> explainGitError e
+      Right _ -> pure ()
--- a/code/hsec-tools/app/Command/Reserve.hs
+++ b/code/hsec-tools/app/Command/Reserve.hs
@@ -1,0 +1,54 @@
+{-# LANGUAGE LambdaCase #-}
+
+module Command.Reserve where
+
+import Control.Monad (unless)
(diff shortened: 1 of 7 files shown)


code/hsec-tools/app/Command/Reserve.hs
```diff
@@ -0,0 +1,63 @@
...
+  , isSecurityAdvisoriesRepo
+  , getGreatestId
+  )
+
+data AssignFlag = AssignId | DoNotAssignId
+
+data CommitFlag = Commit | DoNotCommit
+  deriving (Eq)
+
+go :: Maybe FilePath -> AssignFlag -> CommitFlag -> IO ()
```
blackheaven, line 33:
```suggestion
createReservationFile :: Maybe FilePath -> AssignFlag -> CommitFlag -> IO ()
```

code/hsec-tools/app/Main.hs
```diff
@@ -31,12 +33,31 @@ cliOpts = info (commandsParser <**> helper) (fullDesc <> header "Haskell Advisor
...
         <> command "generate-index" (info commandGenerateIndex (progDesc "Generate an HTML index"))
         <> command "help" (info commandHelp (progDesc "Show command help"))
         )
 
+commandReserve :: Parser (IO ())
+commandReserve =
+  Command.Reserve.go
+  <$> optional (argument str (metavar "REPO"))
+  <*> flag
+        Command.Reserve.DoNotAssignId -- default value
```
TristanCacqueray, line 48: Perhaps `assign` could be the default?

TristanCacqueray (approved):

code/hsec-tools/app/Main.hs
blackheaven: I'm not sure to see the use case to not have it enabled, can you give an example please?

frasertweedale: Eventually, we want people to submit advisories (and reservations) as `HSEC-0000-0000.md`, and have the PR automation assign the IDs.  That is why `--assign` is not the default.

frasertweedale: I would instead propose that the contributor guide, when updated to mention this tool, recommends to use `--assign`.  Then, when the PR automation we desire has been established, we can remove the flag from the contributor how-to doc.

blackheaven: I think it'll make more sense in another command, a `reserve-id` with "no ID" by default does not seem natural to me.

frasertweedale: The process we are (eventually) aiming for is that contributor PRs would use placeholder IDs, and the automation will take care of assigning the appropriate IDs.  In this workflow the preferred user action is to use `hsec-tools` to create the placeholder reservations.  (Similarly with actual advisories - they will submit PR using placeholder IDs, and the automation will assign appropriate IDs).

So, maybe we just need a better subcommand name or CLI options.  For example, instead of *any* default we could require user to specify one of `--placeholder` or `--assign`  Then, as our tooling and automation evolve, we only need to update the documentation as to which option users should use in the typical workflow.

blackheaven: It sounds better.

Maybe:

```
new --placeholder
new --reserved
```

I not sure what would be better `new`, `new-advisory`, `create`, `create-advisory`, any opinion on that?

TristanCacqueray (approved):

code/hsec-tools/app/Main.hs
frasertweedale: I thought about it some more.  Instead of multiple flags, we could have a single option:

```
hsec-tools reserve --id-mode={placeholder|auto}
```

For now there would be no default, meaning user must specify one of the values.  Later when the workflow has the desired automations, we might make `placeholder` the default.

Very much open to other ideas for the command, option and option value names.

frasertweedale: As for subcommand names, I like `new`.

blackheaven: I like this design

frasertweedale:
Updated to implement latest design proposal:

```
Usage: hsec-tools reserve [REPO] --id-mode {placeholder|auto} [--commit]
```

code/hsec-tools/app/Command/Reserve.hs
frasertweedale: changed it to `runReserveCommand`

code/hsec-tools/app/Main.hs
frasertweedale: PR updated with this design.

blackheaven (approved):

PR #116, opened 2023-08-23
Handle CWE name

TristanCacqueray:
This PR enables using CWE names instead of their number to improve advisory visibility.


TristanCacqueray:
If that looks good to you, I'll update the other advisories.

frasertweedale:
Tristan, thank you very much for this.

I have not reviewed in detail yet, but my first and most important comment is: I think we should continue using the CWE number in the TOML.  Why?

- CWE text may change; numbers will not.
- Other databases and system are likely to use the ID as the main index, not the textual description.

That said, our tooling (as we continue to develop and evolve it) should use this new *cwe* library to provide the string representation wherever it presents the data to human users, for example:

- In the HTML index
- In GitHub issues, e.g. a bot that adds a comment or updates the PR description to "summarise" the submission, including an explanation of the CVSS score, CWE descriptions, et cetera.

For the best of both worlds, perhaps we could update the schema to have a list of CWEs that *optionally* include the description, for example:

```
cwe = [
  "328: Use of weak hash",
  "400: Uncontrolled Resource Consumption",
]
```

But the string is optional, and ignored (or perhaps merely checked that it matches the CWE number).

Let me know what you think.  We can discuss at the next meeting.

code/cwe/RenderCsvData.hs
```diff
@@ -0,0 +1,45 @@
...
+module Main where
+
+import Data.List
+import Data.Maybe
+import Text.CSV
+import Text.Read
+
+main :: IO ()
+main = do
+    dbs <- traverse readCSV ["699.csv", "1003.csv"]
```
frasertweedale, line 18: These seem to be CVE views, which are incomplete.
I think we want the full list, which is at the top of the download page.  Especially because we will publish this library for others to use.  Unfortunately there does not seem to be CSV artifact for the full list - only the XML.

code/cwe/src/CWE.hs
```diff
@@ -0,0 +1,26 @@
...
+mkCWEID :: (Integral a, Bits a) => a -> Maybe CWEID
+mkCWEID num = CWEID <$> toIntegralSized num
+
+-- | A map to lookup CWE names.
+cweNames :: Map CWEID Text
+cweNames = Map.fromList (coerce cweData)
+
+-- | A map to lookup CWEID.
+cweIds :: Map Text CWEID
+cweIds = Map.fromList $ (\(k, v) -> (v, k)) <$> (coerce cweData)
```
frasertweedale, line 26: Eventually I would like to add all the parent / child / category membership data as well.  But this is good for now.

frasertweedale (changes requested):

code/hsec-tools/src/Security/Advisories/Parse.hs
```diff
@@ -417,9 +418,10 @@ onKey ::
 onKey tbl k absent present =
   maybe absent present $ Map.lookup k tbl
 
-isInt :: TOML.Value -> TableParser Integer
-isInt (TOML.Integer i) = pure i
-isInt other = throwError $ InvalidFormat "Integer" (describeValue other) "42"
+isCWE :: TOML.Value -> TableParser CWE.CWEID
+isCWE (TOML.Integer i) | Just cwe <- CWE.mkCWEID i, Map.member cwe CWE.cweNames = pure cwe
+isCWE (TOML.String txt) | Just cwe <- Map.lookup txt CWE.cweIds = pure cwe
+isCWE other = throwError $ InvalidFormat "Integer" (describeValue other) "42"
```
frasertweedale, line 424: I think we should continue using the CWE number in the TOML, with optional description text, e.g.:

```
cwe = [
  "328: Use of weak hash",
  400,
]
```

If the summary text is given we should ensure it matches the CWE number (and fail parse if it does not).  But I prefer to leave the CWE ID in the TOML because it makes it easier for humans to look up references.

We could also consider accepting the bare CWE ID as a string, e.g. `"400"`, as a convenience.

code/cwe/src/CWE.hs
```diff
@@ -0,0 +1,26 @@
...
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings #-}
+module CWE (CWEID, mkCWEID, cweNames, cweIds) where
+
+import CWE.Data
+import Data.Text (Text)
+import Data.Coerce
+import Data.Map.Strict as Map
+import GHC.Bits
```
frasertweedale, line 11: Could we please import `Data.Bits` instead?

blackheaven (changes requested):

blackheaven (approved):

code/cwe/RenderCsvData.hs
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env cabal
+{- cabal:
+build-depends: base, xml
+-}
```
blackheaven, line 4: That's unusual

blackheaven (approved):

code/cwe/RenderCsvData.hs
TristanCacqueray: Since it's just for one-shot codegen I figured it was not necessary to add this to the cabal file.

code/cwe/cwe.cabal
```diff
@@ -0,0 +1,27 @@
...
+name:            cwe
+version:         0.1
+synopsis:        Common Weakness Enumaration database
+description:     Use this library to lookup or validate CWE numbers.
+license:         BSD-3-Clause
+author:          Tristan de Cacqueray
+maintainer:      tdecacqu@redhat.com
+category:        Data
+extra-doc-files: CHANGELOG.md
+tested-with:     GHC ==8.10.7 || ==9.0.2 || ==9.2.7 || ==9.4.5 || ==9.6.2
```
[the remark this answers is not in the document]
TristanCacqueray: This PR is adding the lib to the test cabal.project (like for cvss), but I guess we also need to update haskell-ci patch?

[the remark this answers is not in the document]
TristanCacqueray: Actually this is taken care of automatically.

blackheaven (approved):

blackheaven (approved):

TristanCacqueray:
@frasertweedale I'm tempted to close this PR, though I think it can be useful to quickly lookup the name. What do you think, should I rebase it instead?

mihaimaruseac:
I think we might need to redo this if still needed

PR #118, opened 2023-08-25
Add library to validate and score CVSS string

### LINKED ISSUE #101: CVSS type and support library
## Summary

Implement a CVSS parsing and helper library.  Then update the Advisory library to use the CVSS type.

*Note:* the `cvss` field in some advisories may need adjustments to conform.

TristanCacqueray:
Features:
- [x] A new library to handle CVSS String.
- [x] Validate CVSS with hsec-tools.
- [ ] For new advisory PR, implements an action to describe the CVSS through a comment.

Fixes #101


code/cvss/src/Security/CVSS.hs
```diff
@@ -0,0 +1,289 @@
...
+            Nothing -> Left $ "Unknown value: '" <> Text.pack (show value) <> "'"
+            Just _ -> pure ()
+
+{- | Check for required metric
+
+ >>> validateRequired []
+ Left "Missing \"Attack Vector\""
+-}
+validateRequired :: [Metric] -> Either Text ()
+validateRequired metrics = traverse_ checkRequired allMetrics
```
frasertweedale, line 283: I wonder if we should use [`Data.Validation`](https://hackage.haskell.org/package/validation-1.1.3/docs/Data-Validation.html) so that we can capture multiple errors.
But I'm happy to land the initial PR using `Either`, then experiment with `Validation` later.

code/cvss/src/Security/CVSS.hs
```diff
@@ -0,0 +1,289 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+
+{- | This module provides a CVSS parser and utility functions
+ adapted from https://www.first.org/cvss/v3.1/specification-document
+-}
+module Security.CVSS (
+    -- * Type
+    CVSS,
```
frasertweedale, line 10: I think hiding the data constructor is fine.  Essentially really, to avoid construction of invalid data.  But we *should* export `cvssVersion` and the `CVSSVersion` type; these will be important in many use cases.

And we should also have a function to convert a `CVSS` to a CVSS vector string.  It would be good to have two variants: one that re-serialises the vector in the original order, and one that serialises the vector in the preferred order (see e.g. CVSS 3.1 spec Chapter 6).

frasertweedale (changes requested):
Thanks very much @TristanCacqueray.  Overall looks pretty good, but I have a few comments/requests.

blackheaven (approved):

TristanCacqueray:
It seems like the error is happening because the cabal.project doesn't get picked up. I'll rebase that after the CWE PR.

blackheaven:
I think there are two issues:

* Regenerate Haskell-CI pipeline (IIRC, the job rebuilds `cabal.project`)
* It misses nix (`flake.nix`) and related GitHub action pipeline changes

I can handle it on Friday if you want

blackheaven:
Finally

PR #121, opened 2023-09-23
feature: add query command to check wheter a package/version is affected

blackheaven:
## hsec-tools

## TODO

- [x] Add tests
- [x] Add command helper

### DIFF
--- a/code/hsec-tools/app/Main.hs
+++ b/code/hsec-tools/app/Main.hs
@@ -9,6 +9,8 @@ import Data.Maybe (fromMaybe)
 import Data.Foldable (for_)
 import Data.Functor ((<&>))
 import Data.List (intercalate, isPrefixOf)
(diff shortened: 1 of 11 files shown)


blackheaven:
good to go

TristanCacqueray (approved):

mmhat:
Hi, I implemented some suggestions in a PR against this PR: https://github.com/blackheaven/security-advisories/pull/2
As explained there, the original motivation was a function to query advisories for a specific version, but I got a bit carried away and changed some more stuff. As mentioned there, feel free to use it as a whole or cherry-pick the parts you like. While some of the changes are not really important there is one I'd like to see included here: IMHO `listAffectedBy` (or `listVersionRangeAffectedBy` in my PR - I renamed it) should not use `exitFailure` as it makes it more difficult to work with this function if you use hsec-tools as library.

PR #124, opened 2023-11-06
HSEC-2023-0015: cabal-install uses expired key policy

frasertweedale:
Add advisory for cabal-install key policy expiration check issue.

Heads-up @gbaz @Kleidukos - please review the advisory content and let us know if it needs tweaking.

## Advisory

- [ ] It's not duplicated
- [ ] All fields are filled
- [ ] It is validated by `hsec-tools`

PR #125, opened 2023-11-06
fix(ci): check-advisories cache miss breaks due to step call workflow call

blackheaven:
See #124


TristanCacqueray (approved):

frasertweedale:
This seems to fix the issue.  Thanks!

PR #130, opened 2023-11-25
feat(ci): generate and publish advisories website

blackheaven:


mihaimaruseac (approved):

TristanCacqueray (approved):

TristanCacqueray:
The job worked and we now have a gh-pages branch, though the website doesn't seem to be available yet. Should it be at https://haskell.github.io/security-advisories/ ?

frasertweedale:
@TristanCacqueray it's because the branch name is `generated/gh-pages`, not `gh-pages`.

blackheaven:
It's weird because it seems to be correctly configured, I'll dig into it.

blackheaven:
It's may be due to the fact we don't have a haskell.github.io

frasertweedale:
@blackheaven it is the `gh-pages` branch that gets published.  We are pushing to `generated/gh-pages`.

blackheaven:
Not necessarily, it's configured here https://github.com/haskell/security-advisories/settings/pages

frasertweedale:
Ah right... I had no idea it was configurable!  Thanks for the
clarification.

PR #147, opened 2024-02-04
[chore] some tiny improvements

MangoIV:
- always show help on empty comand
- use `hsubparser`

---

## hsec-tools

- [ x Previous advisories are still valid


MangoIV:
perhaps I will add some more commits, I'm only now discovering the tool :)

blackheaven (approved):

PR #153, opened 2024-03-01
fix: try another action for GitHub Pages deployments

blackheaven:
---

## hsec-tools


mihaimaruseac (approved):

frasertweedale:
That seems to have fixed it.  Thanks @blackheaven !

Issue #166, opened 2024-03-17
offer API to request security advisories


MangoIV:
## Summary

it would be nice if we could serve an api that gives back the advisories that is continuously deployed; currently the only way is cloning the repo or parsing the html of the website; neither is immensely efficient

blackheaven:
I will work on a library for that since other consummers (e.g. hackage/flora) will need such a mechanism

blackheaven:
Thanks, don't worry, I'll pack it into a library.

MangoIV:
Where’s this hosted?

frasertweedale:
I think this was closed by mistake.

MangoIV:
@frasertweedale i unfortunately cannot reopen, would you do that for me? Thanks in advance!

MangoIV:
Perfect, thank you! ❤️

blackheaven:
Actually I think it was fix with `hsec-sync`, but maybe I did not understood the problem correctly.

Do you need a way to have a synchronized local copy, or a list of the published advisories?

MangoIV:
well; it would be good if we wouldn't require a user to `git clone` the repository but instead be able to just send a request to some `API` which can 
- give you an update on an advisory, given its ID
- give you all of the advisories, independent of the repo
If you wish, I can build this, I have become very swift at building servant servers ;)

blackheaven:
I see, yes, `hsec-sync` is currently not the best way to achieve that.

On another hand I'm quite worried about this approach for project check as it would end-up with a lot of back-and-forth with the server, moreover I'm reluctant to have the HF host a service (any managed static file hosting would be better).

I think it's better to:

1. Ensure to have a local copy
2. Query against it

I plan to tackle an archive format (without git, #170), tomorrow and/or on Monday, so it will remove a dependency.

MangoIV:
> On another hand I'm quite worried about this approach for project check as it would end-up with a lot of back-and-forth with the server, moreover I'm reluctant to have the HF host a service (any managed static file hosting would be better).

can you elaborate on that? I don’t understand why it would require a lot of back and forth. 

I think it’s fair if the HF doesn’t want to host something like that, especially given that we’d probably have to think about the trust model of something like that but I would at least strive to get something like that.

blackheaven:
> can you elaborate on that? I don’t understand why it would require a lot of back and forth.

[...]

PR #148, opened 2024-02-06
`cabal-audit`: init

MangoIV:
### usage 

```console
λ nix run github:mangoiv/security-advisories/mangoiv/cabal-audit-osv#cabal-audit --
trying to clone https://github.com/haskell/security-advisories
Cloning into '/tmp/cabal-audit-726d3e9345b766bc'...
remote: Enumerating objects: 172, done.
remote: Counting objects: 100% (172/172), done.
remote: Compressing objects: 100% (129/129), done.
remote: Total 172 (delta 6), reused 114 (delta 1), pack-reused 0
Receiving objects: 100% (172/172), 116.55 KiB | 1.31 MiB/s, done.
Resolving deltas: 100% (6/6), done.


Found advisories:

dependency "base" at version 4.18.1.0 is vulnerable for:
  HSEC-2023-0007 "readFloat: memory exhaustion with large exponent"
  published: 2024-04-23 12:43:30 +1000
  https://haskell.github.io/security-advisories/advisory/HSEC-2023-0007
  No fix version available
  toml, parser, dos

dependency "process" at version 1.6.17.0 is vulnerable for:
  HSEC-2024-0003 "process: command injection via argument list on Windows"
  published: 2024-04-23 12:43:30 +1000
  https://haskell.github.io/security-advisories/advisory/HSEC-2024-0003
  Fix available since version 1.6.19.0
  windows
```

> [!Note]
> if this causes some error wrt a lock file incompatibility, upgrade your nix version ;)

### features implemented 

- query the repository for vulnerabilities 
- provide local repository 
- pretty output 
- run the cabal solver on a project and match vulnerabilities against that 
- propose fix version
- link to security-advisories website 
- playtest

### open features 

- don't only provide the latest fix version but a range of fix versions
- check for the newest package on hackage that is greater than the fix version
- more configuration configure the output 
- json output instead of only pretty stdout 
- use API instead of repository
- more solver options (perhaps offer a special flags to pass on options to cabal)
- more structured monadic code (it's all IO for simplicity)

## other changes 

- gave the nix code a polish 
- gave the devshell a polish 
- introduced pre-commit-hooks.nix (please tell me if that's not wanted)

PR #168, opened 2024-03-19
Introduce hsec-sync (#166)

blackheaven:
I have create a new library/executable to synchronize a local security advisories.

Maybe I should also add a `status` operation, WDYT?

See #166 

Maybe be use in #148 and by flora (/cc @tchoutri).


code/hsec-sync/hsec-sync.cabal
```diff
@@ -0,0 +1,95 @@
...
+license:         BSD-3-Clause
+author:          Gautier DI FOLCO
+maintainer:      gautier.difolco@gmail.com
+
+-- A copyright notice.
+-- copyright:
+category:        Data
+extra-doc-files: CHANGELOG.md
+tested-with:
+  GHC ==8.10.7 || ==9.0.2 || ==9.2.8 || ==9.4.8 || ==9.6.3 || ==9.8.1
```
Kleidukos, line 24: Not sure if it's wise to support so many GHC versions for a CLI tool. I would advise that you stay on the highest supported GHC at all times. Most people will anyway take your release binaries, especially in CI.

code/hsec-sync/src/Security/Advisories/Sync/Git.hs
```diff
@@ -0,0 +1,163 @@
...
+          then Left $ GitError cmd $ GitProcessError status stdout stderr
+          else Right GitRepositoryCreated
+
+updateGitRepository :: Repository -> IO (Either GitError ())
+updateGitRepository repo =
+  D.withCurrentDirectory (repositoryRoot repo) $ do
+    _ <- runGit ["remote", "add", "origin", repositoryUrl repo] -- can fail if it exists
+    (setUrlCmd, setUrlStatus, setUrlStdout, setUrlStderr) <-
+      runGit ["remote", "set-url", "origin", repositoryUrl repo]
+    if setUrlStatus /= ExitSuccess
```
Kleidukos, line 122: I believe that some, if not most, of these early returns can be replaced with `when`. (also you should be using `pure`, for reasons of forward-compatibility.)

MangoIV:
for me this is okay as long as I have a directory or git repository url to point it at.

code/hsec-sync/hsec-sync.cabal
blackheaven: Actually, CLI is more a bonus, I had flora/hackage-server in mind when I have made it.

code/hsec-sync/src/Security/Advisories/Sync/Git.hs
blackheaven: IIUC `when` does not shortcut/early return in `IO`.

Kleidukos: So, for context: they could be replaced by when guards and have the appropriate early return effect if there was a mechanism of exception handling that existed. Very much something I expect from my usual web application business in Haskell so much that I didn't realise I had come to expect the same structure everywhere else without explanation. (I will not move an inch on this matter of pure vs. return but you know I love you dearly nonetheless)

MangoIV: You could just use ExceptT

MangoIV:
```haskell
bla :: IO ()
bla =  print =<< runExceptT do
  when True do 
    throwError "oh no" 
  liftIO $ print "hello"
```

I agree the nesting is not beautiful...

PR #179, opened 2024-04-01
Introduce snapshots to distribute advisories

blackheaven:
---

## hsec-tools

## TBD

- [ ] Design review @frasertweedale 
- [x] Rebase @blackheaven 
- [ ] Setup empty branch @blackheaven 
- [ ] Merge @blackheaven 
- [ ] Enable tests @blackheaven


code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,181 @@
...
+ensureSnapshot s =
+  \case
+    SnapshotDirectoryMissing ->
+      ($> SnapshotRepositoryCreated)
+        <$> overwriteSnapshot s
+    SnapshotDirectoryIncoherent ->
+      ($> SnapshotRepositoryCreated)
+        <$> overwriteSnapshot s
+    SnapshotDirectoryInitialized ->
+      return $ Right SnapshotRepositoryExisting
```
MangoIV, line 112: if this was `ExceptT IO`, it would be less awkward.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,181 @@
...
+
+overwriteSnapshot :: Snapshot -> IO (Either SnapshotError ())
+overwriteSnapshot s =
+  runExceptT $ do
+    let root = snapshotRoot s
+    ensuringPerformed <- liftIO $ try $ ensureEmptyRoot root
+    whenLeft ensuringPerformed $
+      throwE . SnapshotProcessError . DirectorySetupSnapshotArchive
```
MangoIV, line 120: `withExceptT (SnapShotProcessError . DirectorySetupSnapshotArchive)`?

MangoIV: (probably can do something similar for the deeply nested thing below)

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,181 @@
...
+    SnapshotRepositoryStatus (..),
+    snapshotRepositoryStatus,
+  )
+where
+
+import qualified Codec.Archive.Tar as Tar
+import qualified Codec.Compression.GZip as GZip
+import Control.Exception (IOException, try)
+import Control.Lens
+import Control.Monad.Extra (unlessM, whenM)
```
MangoIV, line 27: these are `when` and `unless`, you don't need the `Monad` instance, `Applicative` is enough.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,181 @@
...
+                      "Request failed with " <> show (response ^. responseStatus) <> ": " <> show body
+                    _ ->
+                      "Request failed: " <> show content
+      Right result -> do
+        performed <-
+          liftIO $
+            try $
+              withSystemTempDirectory "security-advisories" $ \tempDir -> do
+                let archivePath = tempDir <> "/snapshot-export.tar.gz"
+                BL.writeFile archivePath $ result ^. responseBody
```
MangoIV, line 143: why not just `responseBody result`? https://hackage.haskell.org/package/http-client-0.7.17/docs/Network-HTTP-Client.html#v:responseBody

code/hsec-tools/src/Security/Advisories/Parse.hs
```diff
@@ -1,595 +1,187 @@
...
+        Either e a ->
+        Either ParseAdvisoryError a
+    firstPretty ctr f = first $ mkPretty ctr f
+
+    mkPretty ::
+        (e -> T.Text -> ParseAdvisoryError) ->
+        (e -> T.Text) ->
+        e ->
+        ParseAdvisoryError
+    mkPretty ctr f x = ctr x $ f x
```
MangoIV, line 117: It is usually nicer to, additionally to a `import Data.Text qualified as T` add an import `import Data.Text (Text)`

code/hsec-tools/test/golden/EXAMPLE_ADVISORY.md.golden
```diff
@@ -42,8 +42,8 @@ Right
                         , affectedVersionRangeFixed = Nothing
                         }
                     ]
-                , affectedArchitectures = Nothing
-                , affectedOS = Nothing
+                , affectedArchitectures = Just []
+                , affectedOS = Just []
                 , affectedDeclarations = []
```
MangoIV, line 47: what is the justification for this golden change.

MangoIV: I think it's actually non-sense, the parse should return `Nothing` here. the fields do not exist.

.github/workflows/nix.yml
```diff
@@ -77,6 +77,28 @@ jobs:
...
+          export GIT_WORK_TREE=$DATA_DIR
+          git read-tree "$REF"
+          git add --all --intent-to-add
+          git diff --quiet && exit
+          git add --all
+          TREE=$(git write-tree)
+          git config user.email security-advisories@haskell.org
+          git config user.name "Haskell Security Response Team"
+          COMMIT=$(git commit-tree "$TREE" -p "$REF" -m "$(date --utc --rfc-3339=seconds) ($GITHUB_SHA)")
+          git push origin $COMMIT:$BRANCH
```
MangoIV, line 101: I think it would be good to have a short comment explaining what this is doing, "generate snapshot data" is only helpful if you already know what it's doing

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,181 @@
...
+ensureEmptyRoot root = do
+  D.createDirectoryIfMissing False root
+
+  whenM (D.doesDirectoryExist $ root </> "advisories") $
+    D.removeDirectoryRecursive $
+      root </> "advisories"
+
+  whenM (D.doesFileExist $ root </> "snapshot.json") $
+    D.removeFile $
+      root </> "snapshot.json"
```
MangoIV, line 159: I think this is a dangerous thing, just removing the files without asking.

MangoIV: especially the `removeDirectoryRecursive`

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,181 @@
...
+  \case
+    SnapshotDirectoryMissingE -> "Snapshot directory is missing"
+    SnapshotIncoherent e -> "Snapshot directory is incoherent: " <> e
+    SnapshotProcessError e ->
+      unlines
+        [ "An exception occurred during snapshot processing:"
+        , case e of
+            FetchSnapshotArchive x -> "Fetching failed with: " <> show x
+            DirectorySetupSnapshotArchive x -> "Directory setup got an exception: " <> show x
+            ExtractSnapshotArchive x -> "Extraction got an exception: " <> show x
```
MangoIV, line 64: `displayException`, not `show`.

code/hsec-sync/test/Spec/SyncSpec.hs
```diff
@@ -8,79 +7,87 @@ import Data.Bifunctor (first)
...
               resultResync @?= Right Updated
+              status snapshot >>= (@?= DirectoryIncoherent)
         ]
     ]
 
-withRepositoryAt :: FilePath -> Repository
-withRepositoryAt root =
-  defaultRepository {repositoryRoot = root}
+withSnapshotAt :: FilePath -> Snapshot
+withSnapshotAt root =
```
MangoIV, line 92: the naming here might be a bit unfortunate; `withSomething` usually denotes a function that is of type `Codensity m Something` or `_ -> Codensity m Something`

code/hsec-tools/src/Security/Advisories/Format.hs
```diff
@@ -0,0 +1,432 @@
+{-# LANGUAGE ApplicativeDo #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
MangoIV, line 8: where's the orphan in this module?

code/hsec-tools/src/Security/Advisories/Generate/Snapshot.hs
```diff
@@ -0,0 +1,76 @@
...
+
+-- * Actions
+
+createSnapshot :: FilePath -> FilePath -> IO ()
+createSnapshot src dst = do
+  let toDstFilePath orig = dst </> drop (length src + 1) orig
+
+  forReserved src $ \p _ -> do
+    createDirectoryIfMissing True $ takeDirectory $ toDstFilePath p
+    putStrLn $ "Copying '" <> p <> "' to '" <> toDstFilePath p <> "'"
```
MangoIV, line 38: `hputStrLn sterror`

code/hsec-tools/src/Security/Advisories/Format.hs
```diff
@@ -0,0 +1,432 @@
...
+  f a
+mergeOobMandatory policy oob k ib sendError =
+  mergeOob policy oob k ib (sendError $ "missing mandatory key: " <> k) pure sendError
+
+-- | A solution to an awkward problem: how to delete the TOML
+-- block.  We parse into this type to get the source range of
+-- the first block element.  We can use it to delete the lines
+-- from the input.
+newtype FirstSourceRange = FirstSourceRange (First SourceRange)
+  deriving (Show, Semigroup, Monoid)
```
MangoIV, line 402: I think we should use `DerivingStrategies`
```suggestion
newtype FirstSourceRange = FirstSourceRange (First SourceRange)
  deriving stock Show
  newtype (Semigroup, Monoid)
```

code/hsec-sync/src/Security/Advisories/Sync.hs
```diff
@@ -10,67 +10,66 @@ module Security.Advisories.Sync
...
+status :: Snapshot -> IO RepositoryStatus
+status s =
+  status' s =<< snapshotRepositoryStatus s
 
-status' :: Repository -> GitRepositoryStatus -> IO RepositoryStatus
-status' repo gitStatus = do
-  case gitStatus of
-    GitDirectoryMissing ->
+status' :: Snapshot -> SnapshotRepositoryStatus -> IO RepositoryStatus
+status' s snapshotStatus = do
```
MangoIV, line 51: the `do` is unneeded here, might as well use `\case` for consistency.

code/hsec-sync/src/Security/Advisories/Sync/Url.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE LambdaCase #-}
```
MangoIV, line 1: if you activate these in every module perhaps consider moving them to the `.cabal` file

MangoIV (commented):
Additionally, we should consider being more careful in pulling in new dependencies, either depends on a whole load of things, admittedly the dependencies are quite low in the dependency tree, usually, but I don't think it's justified here. 

A few more informative comments/ docstrings wouldn't hurt either, to increase readability.

I don't think switching the `toml` parsing library is a good idea at all. e.g. because it [doesn't even support toml 1.0.0](https://github.com/kowainik/tomland/issues/373) or it actually does wrong parses in our case (see my review).

I also think treewide formatting changes in files that get heavily changed should be done in a separate `PR`.

code/hsec-tools/src/Security/Advisories/Format.hs
MangoIV: *actually*, we might also use `DerivingVia` and don't wrap `First` *at all*, if we only use it for `Monoid` and `Semigroup` instances.
```suggestion
newtype FirstSourceRange = FirstSourceRange SourceRange
  deriving stock Show
  deriving (Semigroup, Monoid) via (First SourceRange)
```

code/hsec-tools/src/Security/Advisories/Format.hs
```diff
@@ -0,0 +1,432 @@
...
+        <*> Toml.dimap (map unCAPEC) (map CAPEC) (defaultingArrayOf Toml._Integer "capec")
+        .= amdCAPECs
+        <*> Toml.dimap (map unCWE) (map CWE) (defaultingArrayOf Toml._Integer "cwe")
+        .= amdCWEs
+        <*> Toml.dimap (map unKeyword) (map Keyword) (defaultingArrayOf Toml._Text "keywords")
+        .= amdKeywords
+        <*> defaultingArrayOf Toml._Text "aliases"
+        .= amdAliases
+        <*> defaultingArrayOf Toml._Text "related"
+        .= amdRelated
```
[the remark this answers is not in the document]
blackheaven: I agree, It's a formatting issue

code/hsec-tools/hsec-tools.cabal
```diff
@@ -55,14 +63,16 @@ library
...
     , pandoc-types          >=1.22     && <2
     , parsec                >=3        && <4
     , pathwalk              >=0.3
+    , pretty                >=1.0      && <1.2
     , process               >=1.6      && <1.7
     , safe                  >=0.3
     , text                  >=1.2      && <3
     , time                  >=1.9      && <1.14
-    , toml-parser           ^>=2.0.0.0
+    , tomland               >=1.3      && <1.4
```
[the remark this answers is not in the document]
MangoIV: now it's bidirectional but parses wrongly and doesn't support the format?

code/hsec-tools/src/Security/Advisories/Format.hs
```diff
@@ -0,0 +1,432 @@
...
+        "riscv64" -> pure RISCV64
+        "rs6000" -> pure RS6000
+        "s390" -> pure S390
+        "s390x" -> pure S390X
+        "sh4" -> pure SH4
+        "sparc" -> pure SPARC
+        "sparc64" -> pure SPARC64
+        "vax" -> pure VAX
+        "x86_64" -> pure X86_64
+        other -> Left $ "Invalid architecture: " <> T.pack (show other)
```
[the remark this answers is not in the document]
blackheaven: what kind of refactoring do you have in mind?

code/hsec-sync/src/Security/Advisories/Sync/Atom.hs
```diff
@@ -18,7 +19,7 @@ import qualified Text.RSS.Syntax as FeedRSS
 
 latestUpdate :: String -> String -> IO (Either String UTCTime)
 latestUpdate repoUrl branch = do
-  resultE <- try $ get $ repoUrl </> "commits" </> branch </> "advisories.atom"
+  resultE <- try $ get $ mkUrl [repoUrl, "commits", branch, "advisories.atom"]
```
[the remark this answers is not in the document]
MangoIV: ah, this is an `URL`, sure; I think with these error prone things, it would actually be useful to pull a new library, in contrast to `extra` or `either` which provide trivial combinators; are you confident `</>` is correct?

code/hsec-sync/app/Main.hs
```diff
@@ -53,18 +53,18 @@ commandStatus = go <$> repositoryParser
       putStrLn $
         case result of
           DirectoryMissing -> "Directory is missing"
-          DirectoryEmpty -> "Directory is empty"
+          DirectoryIncoherent -> "Directory is incoherent"
```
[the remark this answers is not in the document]
blackheaven: I did not have a good name in mind.

It is used when there is a directory without `advisories` and/or `snapshot.json`.

Maybe `corrupted`/`dirty`, WDYT?

code/hsec-tools/src/Security/Advisories/Format.hs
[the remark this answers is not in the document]
MangoIV: one of the directions is trivial, it's just obtaining the constructor name and then `toLower`, the other is probably best done similarly, by first converting to a `Set` of lowercase versions of the constructors and then looking that up.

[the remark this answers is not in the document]
MangoIV: This can be done with `Generic`, for example.

code/hsec-sync/app/Main.hs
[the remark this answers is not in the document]
MangoIV: I think describing in more detail what it is used for in the error message and/or as a docstring for the constructor would be a good idea, perhaps?

code/hsec-tools/src/Security/Advisories/Format.hs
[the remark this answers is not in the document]
MangoIV: I think you can fix by using `do` with `BlockArguments` or parens. but again, I think it would be less work to just not use a new library...

code/hsec-sync/src/Security/Advisories/Sync.hs
```diff
@@ -10,67 +10,66 @@ module Security.Advisories.Sync
...
-defaultRepository :: Repository
+defaultRepository :: Snapshot
 defaultRepository =
-  Repository
-    { repositoryUrl = "https://github.com/haskell/security-advisories",
-      repositoryRoot = "security-advisories",
-      repositoryBranch = "main"
+  Snapshot
+    { snapshotRoot = "security-advisories",
+      repositoryUrl = "https://snapshothub.com/haskell/security-advisories",
```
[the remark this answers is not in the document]
blackheaven: I have pasted the wrong url, nice catch! thanks

code/hsec-sync/src/Security/Advisories/Sync/Url.hs
```diff
@@ -0,0 +1,27 @@
...
+[x] </> ('/' : ys) = x : '/' : ys
+[x] </> ys = x : '/' : ys
+(x0 : x1 : xs) </> ys = x0 : ((x1 : xs) </> ys)
+
+ensureFile :: String -> String
+ensureFile =
+  \case
+    "" -> ""
+    "/" -> ""
+    (x:xs) -> x : ensureFile xs
```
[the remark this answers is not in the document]
blackheaven: I agree, but `wreq` use a `String` and it's not performance critical.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
blackheaven: It's a plain `case` with two lines each, not a long procedure with nested conditionals, using `ExceptT` here would be overkill.

code/hsec-sync/src/Security/Advisories/Sync/Url.hs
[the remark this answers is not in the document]
MangoIV: so this is fst . partition, behaviourally, right?

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
blackheaven: I did it to avoid excessive bindings, such as:

```haskell
    unlessM (liftIO $ D.doesFileExist metadataPath) $
      throwE SnapshotDirectoryMissingE
```

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
blackheaven: that's how `wreq` suggest to do it.

code/hsec-tools/src/Security/Advisories/Parse.hs
blackheaven: I can change it but it was in the original code

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
MangoIV: you get an `ExceptT` from `overwriteSnapShot`, if you don't run it and you save a hole lot of `fmap`ing, I don't know how it's overkill, you just need to move the `runExceptT` to this function.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
MangoIV: ah! sure; I normally use the `>>= flip` pattern on these.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
MangoIV: yuh, I realised, I guess if you wanna use that instead of `https://hackage.haskell.org/package/http-conduit-2.3.8.3/docs/Network-HTTP-Simple.html` like in my PR, then this is the only way; I don't know what it does justifies the `lens` dependency, too.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
blackheaven: It's the purpose of overwriting a snapshot, asking to details at this point would add complexity while we only expect to have a function which takes care of everything.

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
MangoIV: but it's a matter of taste

code/hsec-tools/src/Security/Advisories/Generate/Snapshot.hs
blackheaven: It's not an error

MangoIV: I think it’s a fair point that hasufell made, that things that are not supposed to be parsed but are just informative output go to stderr.

code/hsec-tools/hsec-tools.cabal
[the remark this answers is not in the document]
MangoIV: I'm a bit partial to not using that specific library but I guess I could not justify this well enough. sorry for that.

code/hsec-tools/src/Security/Advisories/Parse.hs
MangoIV: Aha! Sure :)

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
MangoIV: Okay, perhaps let me reformulate, is it possible that we ever delete something here that the user might want to keep? If that’s the case, we should make it opt in, I think.

MangoIV: But perhaps I’m also overthinking. ;)

code/hsec-tools/src/Security/Advisories/Generate/Snapshot.hs
frasertweedale: I agree.  Diagnostic output should go to stderr.

frasertweedale: There are other cases below also... but I don't see this as a show-stopper.

code/hsec-tools/hsec-tools.cabal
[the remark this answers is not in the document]
frasertweedale: I understand *toml-parser* provides printing support.  This was one of the advantages @glguy mentioned when proposing we switch to *toml-parser* from *toml-reader*.

Let's look into that before switching up the deps and changing all the code, yet again :)

code/hsec-tools/src/Security/Advisories/Format.hs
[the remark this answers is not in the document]
MangoIV:
```haskell
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE OverloadedLists #-}

import Data.Char (toLower)
import Data.Map.Strict (Map)
import Data.Map.Strict qualified as M
import Data.Set (Set)
import Data.Set qualified as S
import Data.Text (Text)
import Data.Text qualified as T

data Arch
  = AArch64
  | Alpha
  | Arm
  | HPPA
  | HPPA1_1
  | I386
  | IA64
  | M68K
  | MIPS
  | MIPSEB
  | MIPSEL
  | NIOS2
  | PowerPC
  | PowerPC64
  | PowerPC64LE
  | RISCV32
  | RISCV64
  | RS6000
  | S390
  | S390X
  | SH4
  | SPARC
  | SPARC64
  | VAX
  | X86_64
  deriving stock (Eq, Ord, Show, Enum, Bounded)

printArchName :: Arch -> Text
printArchName = T.toLower . T.pack . show

allArchNames :: Map Text Arch
allArchNames = M.fromList [(printArchName arch, arch) | arch <- [minBound .. maxBound]]

parseArchName :: Text -> Maybe Arch
parseArchName = (allArchNames M.!?)
```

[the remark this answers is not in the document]
MangoIV: it looks like it might even be more efficient to not use `Map` here but use `lookup` directly on the `[(Text,Arch)]`

code/hsec-tools/hsec-tools.cabal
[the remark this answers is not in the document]
glguy: I'm happy to help with generating TOML.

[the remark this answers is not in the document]
blackheaven: @glguy can you provide any guidance to generate a valid Toml file?

Because it's like `prettyToml` is not really doing that.

> now it's bidirectional but parses wrongly and doesn't support the format?

IIUC, the difference between the current implementation and the official specification are limited to new syntax (mostly).

I'm not sure it's an issue since all the advisories are still passing.

On another hand, having to write a distinct writer is error prone and time consuming.

code/hsec-tools/src/Security/Advisories/Format.hs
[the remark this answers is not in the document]
blackheaven: Actually I have done a package for that, do you think it's worth pulling it?

https://hackage.haskell.org/package/pure-sum

code/hsec-tools/test/golden/MISSING_AFFECTED.md.golden
```diff
@@ -1,17 +1,60 @@
...
+    , modelDatabaseSpecific = Nothing
+    }
+{
+    "details": "## Title\n",
+    "id": "HSEC-0000-0000",
+    "modified": "1970-01-01T00:00:00Z",
+    "published": "1970-01-01T00:00:00Z",
+    "schema_version": "1.5.0",
+    "summary": "Title"
+}
```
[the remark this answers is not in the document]
blackheaven: nice catch, I'll have a look

code/hsec-tools/hsec-tools.cabal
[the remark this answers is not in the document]
glguy: You'll have to be more specific about the issue you're having with prettyToml. It certainly generate valid TOML in general.

Let me know that trouble you've found.

code/hsec-sync/src/Security/Advisories/Sync/Atom.hs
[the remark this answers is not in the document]
blackheaven: I did not find a proper library, do you have any suggestions?

code/hsec-tools/hsec-tools.cabal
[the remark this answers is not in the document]
blackheaven: ok, I have figured it out (modulo one or two changes), thanks everyone

.github/workflows/nix.yml
blackheaven: I'm not really sure what to add here :/

I take any suggestions.

code/hsec-tools/src/Security/Advisories/Format.hs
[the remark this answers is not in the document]
MangoIV:
```haskell
printArchName :: Arch -> Text
printArchName = T.toLower . T.pack . show

parseArchName :: Text -> Maybe Arch
parseArchName = flip lookup [(printArchName arch, arch) | arch <- [minBound .. maxBound]]
```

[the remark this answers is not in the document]
MangoIV: this is probably all the code you need, so I don't think so, no?

blackheaven:
@frasertweedale FYI, I have revert back to `toml-parser` and properly tested `FrontMatter` rendering

code/hsec-sync/src/Security/Advisories/Sync/Atom.hs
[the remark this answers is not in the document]
Kleidukos: @hasufell: I believe you had to deal with URLs/URIs in the recent past. What would you recommend?

blackheaven:
@frasertweedale can you review it again, so I can move forward (see items in the description)

blackheaven:
I won't lie, it was one of the most painful rebase I had to do.

@frasertweedale good luck :)

frasertweedale:
Seven hour train ride on Wednesday.  Should be enough time, I hope :)

frasertweedale:
Why so much reformatting?  :sob: 

@blackheaven did you reformat using some formatter?  If so, it would be better to tackle that as the first commit, and keep the signal-to-noise ratio high for the subsequent commits.

Also, should we consolidate around a particular formatting tool?  (I'm not really a fan, but I'm even more not a fan of what happened in e72abb5ac60f6b3230054c12b40433ecec648d28 ^_^)

Kleidukos:
(I would advise fourmolu, as it is diff-friendly and reduces the noise of each commit)

frasertweedale:
@blackheaven some of the error formatting undoes Mango's Exception instance work.  I can fix it up if you like.

blackheaven:
Actually it was a mistake, @TristanCacqueray enforced fourmolu, but hls does not take le configuration file in account when formatting.

I'm afraid that fixing reformatting during rebase might be a nightmare.

@frasertweedale either you have some time to revert back Mango's work, or highlight where I have overwritten them.

frasertweedale:
> @frasertweedale either you have some time to revert back Mango's work, or highlight where I have overwritten them.

Yes, I'll make the changes and then we can compare and review.  Expect it tomorrow.  Apart from that, there was just one comment so far.  I've read most of the diffs and the new code LGTM at a high level.  Still need to test, also.

frasertweedale:
> > @frasertweedale either you have some time to revert back Mango's work, or highlight where I have overwritten them.
> 
> Yes, I'll make the changes and then we can compare and review. Expect it tomorrow. Apart from that, there was just one comment so far. I've read most of the diffs and the new code LGTM at a high level. Still need to test, also.

I see that later commits in the series put everything back the way it was.  The commit history is a winding road but all is well at the destination.

I will do some testing.  Unless I find bugs / regressions, I will merge it :)

code/hsec-sync/src/Security/Advisories/Sync.hs
```diff
@@ -1,76 +1,95 @@
...
         Left _ ->
           return DirectoryOutDated
         Right info -> do
-          update <- latestUpdate (repositoryUrl repo) (repositoryBranch repo)
+          update <- latestUpdate $ updatesUrl s
           return $
-            if update == Right (zonedTimeToUTC $ lastModificationCommitDate info)
+            if update == Right (lastModificationCommitDate info)
               then DirectoryUpToDate
               else DirectoryOutDated
```
frasertweedale, line 95: I think there is a problem here:

- `update` will be the date of the last commit to the snapshot branch.
- `lastModificationCommitDate` is from the snapshot metadata, and (if I'm not mistaken) is the date of the most recent advisory modification

Therefore, `update` is always newer than `lastModificationCommitDate`, and this expression always yields `DirectoryOutDated`.

I also think that querying the atom feed for the branch from GitHub is a very awkward way to determine whether there is a newer snapshot to retrieve.

GitHub tarballs are served with an ETag.  Could we store the ETag (on the client side).  When we want to check if a newer snapshot is available, perform a `HEAD` request with the `If-None-Match` header.  If the server responds `304 Not Modified`, the snapshot is fresh.  If `200`, the snapshot is out of date.  I tested this and it works as expected on GitHub.  This approach is also likely to be widely compatible with other HTTP servers, if we end up hosting the snapshot elsewhere.

I also tested `If-Modified-Since` and that **does not work** on GitHub.

code/hsec-sync/src/Security/Advisories/Sync.hs
```diff
@@ -1,76 +1,95 @@
...
+  githubSnapshot
+    "security-advisories"
+    "https://github.com/haskell/security-advisories"
+    "generated/snapshot-export"
+
+githubSnapshot :: FilePath -> String -> String -> Snapshot
+githubSnapshot root repoUrl repoBranch =
+  Snapshot
+    { snapshotRoot = root,
+      snapshotUrl = SnapshotUrl $ ensureFile (mkUrl [repoUrl, "archive/refs/heads", repoBranch]) <> ".tar.gz",
```
frasertweedale, line 40: There's an interesting consequence of using the tarballs as generated by GitHub:
```
% curl --silent -L https://github.com/frasertweedale/security-advisories/archive/refs/heads/generated/snapshot-export.tar.gz  | tar -tz |head -n 5
security-advisories-generated-snapshot-export/
security-advisories-generated-snapshot-export/advisories/
security-advisories-generated-snapshot-export/advisories/hackage/
security-advisories-generated-snapshot-export/advisories/hackage/aeson/
security-advisories-generated-snapshot-export/advisories/hackage/aeson/HSEC-2023-0001.md
```

GitHub automatically adds a directory path.  Because `sync` extracts the tarball as-is, subsequent commands using the same `--snapshot-root` will get confused:

```
% hsec-sync sync -d ROOT -r https://github.com/frasertweedale/security-advisories
Snapshot at "ROOT" from "https://github.com/frasertweedale/security-advisories/archive/refs/heads/generated/snapshot-export.tar.gz"
Snapshot just created

% hsec-sync status -d ROOT -r https://github.com/frasertweedale/security-advisories
Directory is incoherent

% ll ROOT
total 0K
drwxr-xr-x. 1 ftweedal ftweedal 46 Jun 15 17:56 security-advisories-generated-snapshot-export

% ll ROOT/security-advisories-generated-snapshot-export 
total 4K
drwxr-xr-x. 1 ftweedal ftweedal 30 Jun 15 17:56 advisories
-rw-r--r--. 1 ftweedal ftweedal 67 Jun 14 16:15 snapshot.json
```

The leading path component is also dependent on the repo name and branch name. 

One approach might be to remove the leading path component during extraction.  I haven't
tried it yet, but probably something like this:

```haskell
Tar.unpack dir . magic . Tar.read =<< BL.readFile tar
  where
    magic = Tar.foldEntries f Tar.Done Tar.Fail
    f entry z
      | entryIsTheTopLevelDir = z  -- omit it
      | otherwise = Tar.Next (stripEntry entry) z
  stripEntry entry = stuff happens
```

code/hsec-sync/hsec-sync.cabal
```diff
@@ -27,22 +27,29 @@ library
...
+    , either       >=5.0   && <5.1
     , extra        >=1.7   && <1.8
     , feed         >=1.3   && <1.4
     , filepath     >=1.4   && <1.5
     , hsec-core    ^>= 0.2
     , http-client  >=0.7.0 && <0.8
     , lens         >=5.1   && <5.3
-    , process      >=1.6   && <1.7
+    , tar          >=0.5   && <0.7
+    , tar-conduit  >=0.3   && <0.5
```
frasertweedale, line 46: We don't seem to be using tar-conduit.

code/hsec-sync/src/Security/Advisories/Sync.hs
blackheaven: It seems reasonable, I'll rewrite it.

blackheaven:
I have done the changes (fixings root directory and using etag), let me know what you think.

frasertweedale:
@blackheaven thanks mate.  I fly home to Australia tomorrow so I will have time to review it!

blackheaven:
Have a nice flight back!

frasertweedale:
uh @blackheaven, did you push your changes?  I don't see them :confused:

blackheaven:
I forgot and it conflicts everywhere, I'll rebase it (again)

blackheaven:
Here we go

code/hsec-sync/src/Security/Advisories/Sync/Snapshot.hs
```diff
@@ -0,0 +1,214 @@
...
+              withSystemTempDirectory "security-advisories" $ \tempDir -> do
+                let archivePath = tempDir <> "/snapshot-export.tar.gz"
+                BL.writeFile archivePath $ result ^. responseBody
+                contents <- BL.readFile archivePath
+                let fixEntry e = e { Tar.entryTarPath = fixEntryPath $ Tar.entryTarPath e }
+                    fixEntryPath :: Tar.TarPath -> Tar.TarPath
+                    fixEntryPath p =
+                      fromRight p $
+                        maybe
+                          (Right p)
```
frasertweedale, line 141: This causes the top-level directory/ies and other top-level paths to be created.  It does work *for our use case*, so I'll merge it *as is*.  But I will probably circle back later to tidy it up.

frasertweedale:
I've reviewed and tested it.  Works as expected.  There are some improvements we can make but no show-stoppers.  Sorry for the delay and thanks for your patience!

I pushed one new commit that adds more detail to the HTTP error messages.  I'll merge when green.

frasertweedale:
CI is busted.  Dealing with it in PR https://github.com/haskell/security-advisories/pull/220.

frasertweedale:
Finally merged.  Thanks for implementing this feature, @blackheaven!

PR #186, opened 2024-04-03
atom: omit old entries from feed

frasertweedale:
**NOTE** I don't know if this is a proper thing to do or not.  But it seems intuitive that
an atom feed should not grow unbounded.

To prevent unlimited growth of the feed, omit old entries.  The heuristic is:

1. sort advisories by modification date
2. keep all advisories that were modified in the last 30 days
3. add older advisories until the feed has at least 10 entries

---

## Advisory

- [ ] It's not duplicated
- [ ] All fields are filled

## hsec-tools

### DIFF
--- a/code/hsec-tools/src/Security/Advisories/Generate/HTML.hs
+++ b/code/hsec-tools/src/Security/Advisories/Generate/HTML.hs
@@ -16,7 +16,7 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import qualified Data.Text.Lazy as TL
-import Data.Time (ZonedTime, zonedTimeToUTC)
+import Data.Time (UTCTime, ZonedTime, addUTCTime, getCurrentTime, nominalDay, zonedTimeToUTC)
 import Data.Time.Format.ISO8601
 import System.Directory (createDirectoryIfMissing)
 import System.Exit (exitFailure)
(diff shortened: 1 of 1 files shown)


code/hsec-tools/src/Security/Advisories/Generate/HTML.hs
```diff
@@ -257,11 +258,25 @@ feed advisories =
...
+-- Finally, if there are fewer than 10 such advisories, top up the
+-- list to include 10 entries.
+--
+sortAndDropOld :: UTCTime -> [Advisories.Advisory] -> [Advisories.Advisory]
+sortAndDropOld now =
+  (\(h,t) -> h <> take (10 - length h) t)
+  . span ((> thirtyDaysAgo) . zonedTimeToUTC . Advisories.advisoryModified)
+  . sortOn (Down . zonedTimeToUTC . Advisories.advisoryModified)
+  where
+    thirtyDaysAgo = addUTCTime (negate (nominalDay * 30)) now
```
blackheaven, line 279: I feel that it's quite confusing/redundant.

I use a feed reader and whenever I add a new feed, I have the last 10 articles marked as "unseen", it gives an explicit health status of the feed (i.e. "old entries = not really active" vs "no entries = broken ?")

frasertweedale:
Converting to draft until we work out whether to proceed, or not.

code/hsec-tools/src/Security/Advisories/Generate/HTML.hs
mihaimaruseac: I think so too. In fact, I get upset when I add a feed and only the most recent posts are there and then I have to go back to the source and find out the other posts.

I think we should wait and see if the feed becomes too long that this becomes an issue. We can also alleviate this problem (at that time) by reducing the amount of text included in the feed, instead of the full content we can only include a short summary / start of entry?

frasertweedale: Fair enough; thanks for the feedback.  Consensus is not to land it so I shall close the ticket.  If it becomes an issue later we could offer a "recent" feed in addition to complete feed.

frasertweedale:
Consensus is not to proceed.  Closing the PR.

PR #188, opened 2024-04-09
HSEC-2024-0003: process: command injection on Windows

TristanCacqueray:
## Advisory

- [x] It's not duplicated
- [x] All fields are filled

### DIFF
--- a/advisories/hackage/process/HSEC-2024-0003.md
+++ b/advisories/hackage/process/HSEC-2024-0003.md
@@ -1,0 +1,134 @@
+```toml
+[advisory]
+id = "HSEC-2024-0003"
+cwe = [150]
+keywords = ["windows"]
+aliases = ["VU#123335"]
+related = ["CVE-2024-1874", "CVE-2024-24576", "CVE-2024-22423"]
+
+[[references]]
+type = "ARTICLE"
+url = "https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/"
+
+
+[[affected]]
+package = "process"
+os = ["mingw32"]
+cvss = "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H"
+
+[[affected.versions]]
+introduced = "1.0.0.0"
+fixed = "1.16.19.0"
+```
+
+# process: command injection via argument list on Windows
+
+The *process* library on Windows is vulnerable to a command injection
+vulnerability, via `cmd.exe`'s interpretation of arguments.  Processes that
+invoke batch files (`.bat`, `.cmd`) and pass arguments whose values are
+affected by program inputs may be affected.
+
+This issue was discovered in many programming languages' Windows process
+execution behaviour.  It was tracked by CERT/CC as **VU#123335** and a
+coordinated disclosure was made on 2024-04-09 17:00 UTC.
+
+
+## Background
+
+Unlike POSIX systems, Windows does not have a mechanism for passing multiple
+arguments.Command line parsing is up to individual programs.
+
+The *process* library defines the `RawCommand` constructor for specifying an
+executable and its arguments:
+
+```haskell
+data CmdSpec
+  = ShellCommand String
+  | RawCommand FilePath [String]
+```
+
+On Windows, the `RawCommand` executable name and arguments are serialised into
+a single *command line* string, with separate arguments quoted separately.
+*process* then invokes the Windows [`CreateProcess`][doc-CreateProcess]
+routine with this command line string is given as the `lpCommandLine`
+argument.
+
+
+## Issue
+
+When executing `.bat` or `.cmd` files, [`CreateProcess`][doc-CreateProcess]
+implicitly spawns `cmd.exe`.  The `System.Process` command line construction
+does not escape characters with special meaning to `cmd.exe`.  As a
+consequence, a command injection vulnerability arises when the following
+conditions are satisfied:
+
+- Program running on Windows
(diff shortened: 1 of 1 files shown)


blackheaven (approved):

bgamari:
`process-1.6.19.0` has been [released](https://hackage.haskell.org/package/process).

TristanCacqueray:
Great, thank you @bgamari !

frasertweedale:
Whoops, I typo'd the fix version as `1.16.19.0` instead of `1.6.19.0`.  I just pushed a commit to fix that, directly to `main`.

hasufell:
> `process-1.6.19.0` has been [released](https://hackage.haskell.org/package/process).

So this release violates PVP?

frasertweedale:
@hasufell see discussion at https://github.com/haskell/security-advisories/blob/main/advisories/hackage/process/HSEC-2024-0003.md#fix-versions.  The scope of behavioural change is as restricted as possible: https://github.com/haskell/process/commit/3c419f9eeedac024c9dccce544e5a6fb587179a5#diff-eca1a11435b6f39e5c276338597973782ee72e6d17d2aa429df113f817aa224a

hasufell:
I find the reasoning questionable.

Did we just admit that PVP is a failure when it comes to security updates?

frasertweedale:
@hasufell https://pvp.haskell.org/ doesn't even talk about behavioural changes, only API changes.  But morally, I agree that potentially breaking behavioural changes would *usually* warrant a major release but...

A program (or user) may rely somehow on buggy behaviour.  Can we never fix a bug without bumping the major version?

Especially when it comes to security issues, it is appropriate to balance the likelihood of a behavioural change breaking something, and the ease of adoption by users (major version bumps cause more pain or toil).

PVP is an important guideline and valuable tool.  A carefully considered deviation from (a particularly interpretation of) it does not mean it is a failure.

bgamari:
@hasufell as far as I can tell the PVP does not give a clear prescription here. The PVP's MUST prescriptions are all to do with the overall shape of the exposed interfaces (that is, names, types, and children) and orphan instances.

However, my initial inclination was very similar to yours; it feels wrong to change the behavior of a function in a minor release. However, after discussion, @tomjaguarpaw and I came down on versioning the new release with a minor bump; this determination was made in light of the facts that:

 * the change affects only a small number of cases (`.bat` and `.cmd` files on Windows)
 * the new behavior should be functionally equivalent to the old behavior in all but a very small number of cases

These considerations were weighed against the fact that a major bump would have made it extremely difficult for us to ensure that this change would be made available in future minor GHC releases. Afterall, we have historically maintained a policy of avoiding major version bumps in core libraries in such releases. This policy is what ensures that users can readily migrate to new minor GHC releases (and benefit from the correctness and security fixes that they bring).

hasufell:
> These considerations were weighed against the fact that a major bump would have made it extremely difficult for us to ensure that this change can be distributed with future minor GHC releases. Afterall, we have historically maintained a policy of avoiding major version bumps in core libraries in such releases. 

Sure. What I mean is this: what if the next time a security bugfix in a boot library requires changing types/exposed API?

Something doesn't seem right. Are we unable to ship such bugfixes swiftly across the Haskell ecosystem? 

This seems like a far reaching problem:

- users depending on ghc the library get locked in (it should be reinstallable)
- cabal users don't get the security update by default: https://github.com/haskell/cabal/issues/9669
- no defacto standard yet to audit your dependencies
- what about stackage?

I'm not so much arguing about the current decision, but it's an opportunity to open this discussion.

frasertweedale:
> Sure. What I mean is this: what if the next time a security bugfix in a boot library requires changing types/exposed API?
> 
> ...
> 
> I'm not so much arguing about the current decision, but it's an opportunity to open this discussion.

Thank you for opening the discussion.

In my experience, security issues *requiring* a breaking API change are uncommon.  What I have sometimes seen is an issue that prompts an API change (e.g. to ensure safer usage of the library) for the future release series, with the issue resolved in some other way that does not break API compatibility for existing supported release branches.

It would be worthwhile to consider what we could/should do in plausible "worst case scenarios", and write them down.  Also to identify barriers that prevent or nullify the ideal or preferred response (you already pointed out some specific problems above).

hasufell:
> In my experience, security issues requiring a breaking API change are uncommon.

Well, I think that mostly applies to C. We don't have a lot of knowledge in Haskell about it.

The most prominent recent-ish security bugfix was aeson-2 and it did break API.

GHCup had a severe bug that could have qualified as a vulnerability (not code execution, but filesystem corruption) and fixed it through breaking API.

I think it's much more common in Haskell to fix unexpected computation through encoding more invariants in the type system.

> It would be worthwhile to consider what we could/should do in plausible "worst case scenarios", and write them down.

Yes, the problem right now is that I as a core library maintainer rely on GHC HQ to sorta ship security bugfixes in time, which I don't feel is a good strategy.

frasertweedale:
@hasufell thanks for reminding me of the aeson situation (and fix).  Could you please point to more info about the GHC filesystem corruption bug?  I wasn't aware of it.

hasufell:
> Could you please point to more info about the GHC filesystem corruption bug? I wasn't aware of it.

That was GHCup.

- https://gitlab.haskell.org/haskell/ghcup-hs/-/merge_requests/253
- https://github.com/haskell/ghcup-hs/blob/master/CHANGELOG.md#01178----2022-05-11
- https://github.com/haskell/ghcup-hs/commit/991e540c11fd25758529ea277d3d61918162b0b1
- https://discourse.haskell.org/t/ann-ghcup-0-1-17-8-bugfix-release/4510

PR #193, opened 2024-04-20
Add guide to secure GitHub workflows

mihaimaruseac:
I'm not sure of where to place this. I made a new directory, but leaving to review to decide if we want this in the root directory or somewhere else. My thinking was that we can add other guides too as we go along, for example when we identify common vulnerable patterns in code.

### DIFF
--- a/guides/github.md
+++ b/guides/github.md
@@ -1,0 +1,121 @@
+# How to secure GitHub repositories
+
+It is recommended to turn on 2FA for most repositories, especially if building
+a package that is a dependency of multiple other packages.
+
+At least for critical packages, administrators should enable branch
+protection. Require CI to pass before merging to the main branch. Allow only
+repository owners to merge PRs.
+
+It is ideal to require PR reviews before merging. For security critical
+packages, having at least two reviewers is ideal, as this alleviates the risk
+of sock-puppet accounts.
+
+To keep dependencies up to date, it is advisable to enable
+[Dependabot][dependabot] or [Renovate][renovate] or a similar tool which will
+create PRs to update dependencies and alert on vulnerabilities in
+dependencies. Both of these tools can be configured to send a PR at regular
+intervals (e.g., once a week).
+
+It is recommended to install [Scorecards action][scorecard] for public
+repositories and trying to improve the score as high as possible. This is a
+scanner for security best practices, most of which are already discussed in
+this document.
+
+## Securing GitHub Actions
+
+It is preferable to use the hosted runners as possible. The large runners
+supported by GitHub come with usage costs, but are better from the point of
+view of supply chain security than hosting your own infrastructure for CI.
+
+All GitHub Actions workflows should restrict [permissions][gha-permissions] to
+the minimum scope needed. Scope the permissions at job level, instead of at
+workflow level.
+
+```
+  check-tests:
+    permissions:
+      actions: read
+```
+
+Minimize usage of actions that create PRs or push code to branch. Thoroughly
+inspect actions that can approve PRs and workflows that are triggered after a
+PR has been approved (term-of-check-vs-term-of-use type of concerns).
+
+If using actions outside of your organisation, these should be pinned by
+commit hash. Don't pin be version tag as these tags can be forced pushed. The
+[Scorecards Pinned Dependencies workflow][scorecard] can help identify which
+actions should still be pinned. [Dependabot][dependabot] is able to upgrade
+the versions of these actions periodically:
+
+```
+      - name: Upload sdist as an artifact for later jobs in workflow
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          ...
+```
+
+If secrets are used as arguments to commands for the workflows, pass them via
+environment variables. The action log can leak command line arguments!
+
+If using secrets, these should be scrubbed as soon as they are no longer
+needed. Make sure they don't show up in logs, including debug runs of the
+workflow!
+
+Be extra careful of GitHub context variables (e.g.,
+`github.event.issue.title`) which are vulnerable to command injection attacks.
+Don't use these directly in commands. Prefer passing them via environment
+variables. Use [Scorecards][scorecard] to identify dangerous workflows.
+
+Audit carefully all workflows of type `pull_request_target` and
+`workflow_run`. For these, don't use secrets at all and separate the sensitive
+parts into separate runs. Validate all inputs and only run these workflows on
+protected branches. Never use `pull_request_head.ref` as the branch head
+reference since this can be updated between the time the PR gets approved and
+the time the CI workflow starts executing.
+
+### Using self-hosted runners
+
+If using self-hosted runners, create 2 separate pools: one pool to be used
+only for CI for PRs and another one to be used for releases and other
+sensitive jobs. Ideally, don't share these pools across multiple projects in
+the same organisation.
+
+> [!WARNING]
+> Repository-level self-hosted runners are accessible to any workflow in the
+> repository. Never trigger a CI job before the PR is reviewed as a PR can
+> maliciously alter the workflow to attack the hosted runner.
+
+Don't allow any secrets in jobs that run in the CI pool. Only use secrets on
+the release pool and never trigger jobs on the release pool on code that is
+not reviewed.
+
+Ideally, the self-hosted runners should be ephemeral. Destroy them as soon as
+the CI run finishes. If caching between jobs is required, periodically recycle
+all runners (e.g., destroy them every week).
+
+Do not allow writing to the repository from the workflow actions if using
+self-hosted runners. A compromise of the runner can result in compromise of
+the repository otherwise.
+
+All CI runners should have a timeout. All workflows that run on CI runners
+should not run on forks of the repository. Gate workflows run by checking the
+value of the `github.repository` variable:
+
+```
+jobs:
+  build:
+    if: github.repository == 'haskell/haskell-language-server' # Don't do this in forks
+```
+
+You can also use `pull_request.author_association` to differentiate between
+owners of the project and collaborators added to the project that also have
+write access.
+
+Monitor all activity on the runners and have separate infrastructure to scan
+for malicious activity.
+
+[scorecard]: https://github.com/ossf/scorecard-action
+[dependabot]: https://github.com/dependabot
+[renovate]: https://github.com/renovatebot/renovate
+[gha-permissions]: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
--- a/guides/github.md
+++ b/guides/github.md
@@ -7,6 +7,15 @@ At least for critical packages, administrators should enable branch
 protection. Require CI to pass before merging to the main branch. Allow only
 repository owners to merge PRs.
 
+> [!WARNING]
+> It is recommended to run workflows only after the PR has been reviewed. The
+> "Require approval for all outside collaborators" setting is the recommended
+> one. However, this can cause friction in repositories with a high number of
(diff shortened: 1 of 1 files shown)


blackheaven (approved):

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+> "Require approval for all outside collaborators" setting is the recommended
+> one. However, this can cause friction in repositories with a high number of
+> PRs but small number of contributors. In this case, if all contributors are
+> trusted, "Require approval for first-time contributors" is a valid option,
+> but dedicate more efforts to separate sensitive steps into workflows that
+> can only run after approval.
+
+It is ideal to require PR reviews before merging. For security critical
+packages, having at least two reviewers is ideal, as this alleviates the risk
+of sock-puppet accounts.
```
frasertweedale, line 21: In case someone doesn't know what "sock puppet account" means, perhaps link to https://en.wikipedia.org/wiki/Sock_puppet_account ?

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+intervals (e.g., once a week).
+
+It is recommended to install [Scorecards action][scorecard] for public
+repositories and trying to improve the score as high as possible. This is a
+scanner for security best practices, most of which are already discussed in
+this document.
+
+## Securing GitHub Actions
+
+It is preferable to use the hosted runners as possible. The large runners
```
frasertweedale, line 36: "hosted runners **whenever** possible" ?

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+
+```
+  check-tests:
+    permissions:
+      actions: read
+```
+
+Minimize usage of actions that create PRs or push code to branch. Thoroughly
+inspect actions that can approve PRs and workflows that are triggered after a
+PR has been approved (term-of-check-vs-term-of-use type of concerns).
```
frasertweedale, line 52: "**time**-of-check-vs-**time**-of-use"

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+
+If using actions outside of your organisation, these should be pinned by
```
frasertweedale, line 54: Perhaps

> "If using actions **defined** outside your org..."

or something similar.

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+commit hash. Don't pin be version tag as these tags can be forced pushed. The
```
frasertweedale, line 55: > Don't pin **by** version tag

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+Ideally, the self-hosted runners should be ephemeral. Destroy them as soon as
+the CI run finishes. If caching between jobs is required, periodically recycle
+all runners (e.g., destroy them every week).
+
+Do not allow writing to the repository from the workflow actions if using
+self-hosted runners. A compromise of the runner can result in compromise of
+the repository otherwise.
+
+All CI runners should have a timeout. All workflows that run on CI runners
+should not run on forks of the repository. Gate workflows run by checking the
```
frasertweedale, line 111: > Gate **workflow runs** ...

frasertweedale (commented):
Thanks, great resource.  Only some editorial suggestions.

guides/github.md
```diff
@@ -0,0 +1,130 @@
+# How to secure GitHub repositories
+
+It is recommended to turn on 2FA for most repositories, especially if building
+a package that is a dependency of multiple other packages.
```
hasufell, line 4: - 2FA can be easily circumvented with apps that understand TOTP, such as keepassxc
- 2FA apps are often installed on mobile devices, which are among the most insecure devices available

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+
+At least for critical packages, administrators should enable branch
+protection. Require CI to pass before merging to the main branch. Allow only
```
hasufell, line 7: I don't understand how branch protection improves *security*. Attacks on github repos have been carried out previously through forks. If the package is on hackage, it is easy to recover the code.

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+> trusted, "Require approval for first-time contributors" is a valid option,
+> but dedicate more efforts to separate sensitive steps into workflows that
+> can only run after approval.
+
+It is ideal to require PR reviews before merging. For security critical
+packages, having at least two reviewers is ideal, as this alleviates the risk
+of sock-puppet accounts.
+
+To keep dependencies up to date, it is advisable to enable
+[Dependabot][dependabot] or [Renovate][renovate] or a similar tool which will
```
hasufell, line 24: Both tools don't support Haskell afaik, so why are they recommended?

- https://github.com/dependabot/dependabot-core/issues/2745
- https://github.com/renovatebot/renovate/issues/8187

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+It is recommended to install [Scorecards action][scorecard] for public
+repositories and trying to improve the score as high as possible. This is a
+scanner for security best practices, most of which are already discussed in
+this document.
+
+## Securing GitHub Actions
+
+It is preferable to use the hosted runners as possible. The large runners
+supported by GitHub come with usage costs, but are better from the point of
+view of supply chain security than hosting your own infrastructure for CI.
```
hasufell, line 38: What is the reasoning for this claim?

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+the minimum scope needed. Scope the permissions at job level, instead of at
+workflow level.
+
+```
+  check-tests:
+    permissions:
+      actions: read
+```
+
+Minimize usage of actions that create PRs or push code to branch. Thoroughly
```
hasufell, line 50: What about tools/bots like mergify, which is used by cabal?

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+protected branches. Never use `pull_request_head.ref` as the branch head
+reference since this can be updated between the time the PR gets approved and
+the time the CI workflow starts executing.
+
+### Using self-hosted runners
+
+If using self-hosted runners, create 2 separate pools: one pool to be used
+only for CI for PRs and another one to be used for releases and other
+sensitive jobs. Ideally, don't share these pools across multiple projects in
+the same organisation.
```
hasufell, line 91: What is the reasoning? Github can be configured to require approval before workflows are run from outside contributors. This attack vector is already covered.

guides/github.md
```diff
@@ -0,0 +1,130 @@
...
+> [!WARNING]
+> Repository-level self-hosted runners are accessible to any workflow in the
+> repository. Never trigger a CI job before the PR is reviewed as a PR can
+> maliciously alter the workflow to attack the hosted runner.
+
+Don't allow any secrets in jobs that run in the CI pool. Only use secrets on
+the release pool and never trigger jobs on the release pool on code that is
+not reviewed.
+
+Ideally, the self-hosted runners should be ephemeral. Destroy them as soon as
```
hasufell, line 102: Can you explain how this is supposed to work?

guides/github.md
mihaimaruseac: Thank you. Added to #194

guides/github.md
mihaimaruseac: Thank you. Added to #194

guides/github.md
mihaimaruseac: Thank you. Added to #194

guides/github.md
mihaimaruseac: Thank you. Added to #194

guides/github.md
mihaimaruseac: Thank you. Added to #194

guides/github.md
mihaimaruseac: Thank you. Added to #194

guides/github.md
mihaimaruseac: That is true, but having 2FA adds an additional hurdle for attackers. I added a few more lines to include this comment

guides/github.md
mihaimaruseac: I was wrong. I thought the support already existed. I can remove them and add them later when we have support.

But I'd still like to have a section about updating dependencies and I would still recommend these for updating github actions (and other non-Haskell packages).

For now, I'm going to add a mention that support for Haskell is missing at the moment.

I will also recommend https://github.com/nomeata/haskell-bounds-bump-action

guides/github.md
mihaimaruseac: You no longer need to take steps in securing your own infrastructure. It's a trade-off of costs between what you'd spend on securing your infra vs what you'd spend on paying GitHub

guides/github.md
mihaimaruseac: I think if you thoroughly review them and are ok with what they do, they are ok. It is not a "you shall not use these actions" but more of a "please review these before using them"

guides/github.md
mihaimaruseac: Sure, if you approve all PRs before running these actions then the vector is covered. But many people consider that asking for review before any CI runs for everyone is too much. So they either pick the middle ground and require approval only for the first contribution (which then is bypassed by a typo fix) or add the contributor as a collaborator to the project.

Plus, if you missed something during review, or the attacker manages to exploit a TOCTOU issue, they could still poison the runners.

guides/github.md
mihaimaruseac: One possibility is to have an image of the runner in the cloud account, then use a REST API or similar to trigger creating a new VM based on that image (see https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#create-configuration-for-a-just-in-time-runner-for-an-organization for example). This image stays up for the duration of CI and then gets deleted.

mihaimaruseac:
@hasufell thank you for the thorough review!

guides/github.md
hasufell: > The xz attack was possible because Jia Tan was able to write to the repository directly

That's a matter of access, not of branch protection rules.

You can have secure access policies without any branch protection rules.

I'd even argue it's detrimental to security if you:

- protect master branch
- give all sorts of people write access to non-protected branches 

It's preferable to interact with contributors through forks.

guides/github.md
hasufell: This sounds like a rather bold assumption that Github has more secure infrastructure than some machine on managed amazon cloud.

guides/github.md
hasufell: But the text is only mentioning CI actions, not such github apps/bots.

guides/github.md
hasufell: I think it would be good to include that link

guides/github.md
mihaimaruseac: (sorry for coming late to this, was swamped by other tasks at work and continuity tasks in face of layoffs)

I partially agree with you. Yes, interacting with contributors via forks is ideal. But having branch protection also protects against accidental pushes to master. By branch protection here all I mean is that merging to master should be done by approved users, only after CI passes, from PRs and not direct pushes.

I'll make a new PR to expand on this

guides/github.md
mihaimaruseac: It's more an assumption on the fact that GitHub has a security team that spent a lot of time to look at securing infrastructure. Of course if someone spends time to secure their own infra to what they think is a better security standard, they should use their own infra.

A student that just starts a new project? Better to use Github's CI that their own. A big company with dedicated security team and cloud infra? Better to use their own CI

guides/github.md
mihaimaruseac: It is mentioned in the first warning box in the doc: https://github.com/haskell/security-advisories/blob/main/guides/github.md

guides/github.md
mihaimaruseac: I'll add it to the next PR

PR #201, opened 2024-05-31
[fix] some minor fixes

### LINKED ISSUE #200: add  `Exception` instance to  to `ParseAdvisoryError`
## Summary

I would like to use the `displayException` function such that I do not have to reimplement this 

```
169   │         case e of
170   │           MarkdownError _ explanation -> "Markdown parsing error:\n" <> explanation
171   │           MarkdownFormatError explanation -> "Markdown structure error:\n" <> explanation
172   │           TomlError _ explanation -> "Couldn't parse front matter as TOML:\n" <> explanation
173   │           AdvisoryError _ explanation -> "Advisory structure error:\n" <> explanation
```

I wonder whether you could add a default instance for `Exception` where you override `displayException` to do what the code in `hsec-tools/app/Main.hs` does.

Thanks in advance!

### LINKED ISSUE #202: No ISO 8601 parse for outputs of git into type `ZonedTime`
## Summary

downstream may have issues parsing formats like this output by git `2024-05-30T16:15:18Z` into `ZonedTime` 

This makes oob timestamp parsing fail. the error is not printed, that's why the error is the most confusing

the reason for this behavior is that hsec-tools parses into `ZonedTime`. git may not output ZonedTime.

### LINKED ISSUE #203: drop `ZonedTime` in favour of `UTCTime`
## Summary

This issue exists for evaluating the advantages and disadvantages of dropping `ZonedTime` in favor of `UTCTime`. The main reason I want this is that `ZonedTime` doesn't admit an `Ord` instances making it hard to use e.g. with `containers`. Additionally, not all timestamps output by `git` or parsed from toml are zoned, hence we're inventing a bunch of data which doesn't have to be the case. For me the question is what do we actually need the time zones for? 

Blocked by https://github.com/glguy/toml-parser/issues/19

MangoIV:
- introduce Exception instance for ParseAdvisoryError
- display if parsing out of band git attributes failed
- also try to parse into a UTCTime **instead** of a ZonedTime

resolves #203 
resolves #202 
resolves #200 

---

## hsec-tools

### DIFF
--- a/code/hsec-tools/app/Main.hs
+++ b/code/hsec-tools/app/Main.hs
@@ -1,13 +1,14 @@
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE BlockArguments #-}
 
 module Main where
 
 import Control.Monad (forM_, join, void, when)
+import Control.Monad.Trans.Except (runExceptT, ExceptT (ExceptT), withExceptT, throwE)
+import Control.Monad.IO.Class (liftIO)
 import qualified Data.ByteString.Lazy as L
 import Data.Maybe (fromMaybe)
 import Data.Foldable (for_)
-import Data.Functor ((<&>))
 import Data.List (intercalate, isPrefixOf)
 import Distribution.Parsec (eitherParsec)
 import Distribution.Types.VersionRange (VersionRange, anyVersion)
@@ -154,15 +155,14 @@ withAdvisory :: (Maybe FilePath -> Advisory -> IO ()) -> Maybe FilePath -> IO ()
 withAdvisory go file = do
   input <- maybe T.getContents T.readFile file
 
-  oob <- ($ emptyOutOfBandAttributes) <$> case file of
-    Nothing -> pure id
-    Just path ->
-      getAdvisoryGitInfo path <&> \case
-        Left _ -> id
-        Right gitInfo -> \oob -> oob
-          { oobPublished = Just (firstAppearanceCommitDate gitInfo)
-          , oobModified = Just (lastModificationCommitDate gitInfo)
-          }
+  oob <- runExceptT case file of
+    Nothing -> throwE StdInHasNoOOB
+    Just path -> withExceptT GitHasNoOOB do 
+      gitInfo <- ExceptT $ liftIO $ getAdvisoryGitInfo path 
+      pure OutOfBandAttributes
+        { oobPublished = firstAppearanceCommitDate gitInfo
+        , oobModified = lastModificationCommitDate gitInfo
+        }
 
   case parseAdvisory NoOverrides oob input of
     Left e -> do
--- a/code/hsec-tools/app/Main.hs
+++ b/code/hsec-tools/app/Main.hs
@@ -19,6 +19,7 @@ import Validation (Validation(..))
 import qualified Data.Aeson
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
+import Control.Exception (Exception(displayException))
 import Options.Applicative
 
 import Security.Advisories
@@ -165,12 +166,7 @@ withAdvisory go file = do
 
   case parseAdvisory NoOverrides oob input of
     Left e -> do
-      T.hPutStrLn stderr $
-        case e of
-          MarkdownError _ explanation -> "Markdown parsing error:\n" <> explanation
-          MarkdownFormatError explanation -> "Markdown structure error:\n" <> explanation
-          TomlError _ explanation -> "Couldn't parse front matter as TOML:\n" <> explanation
-          AdvisoryError _ explanation -> "Advisory structure error:\n" <> explanation
+      hPutStrLn stderr (displayException e)
       exitFailure
     Right advisory -> do
       go file advisory
--- a/code/hsec-tools/app/Main.hs
+++ b/code/hsec-tools/app/Main.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE BlockArguments #-}
 
 module Main where
 
@@ -155,9 +154,9 @@ withAdvisory :: (Maybe FilePath -> Advisory -> IO ()) -> Maybe FilePath -> IO ()
 withAdvisory go file = do
   input <- maybe T.getContents T.readFile file
 
-  oob <- runExceptT case file of
+  oob <- runExceptT $ case file of
     Nothing -> throwE StdInHasNoOOB
-    Just path -> withExceptT GitHasNoOOB do 
+    Just path -> withExceptT GitHasNoOOB $ do 
       gitInfo <- ExceptT $ liftIO $ getAdvisoryGitInfo path 
       pure OutOfBandAttributes
         { oobPublished = firstAppearanceCommitDate gitInfo
--- a/code/hsec-tools/src/Security/Advisories/Filesystem.hs
+++ b/code/hsec-tools/src/Security/Advisories/Filesystem.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BlockArguments #-}
 {-|
 
 Helpers for the /security-advisories/ file system.
@@ -127,7 +126,7 @@ listAdvisories root =
     if isSym
       then return $ pure []
       else do
-        oob <- runExceptT $ withExceptT GitHasNoOOB do
+        oob <- runExceptT $ withExceptT GitHasNoOOB $ do
           gitInfo <- ExceptT $ liftIO $ getAdvisoryGitInfo advisoryPath
           pure OutOfBandAttributes
             { oobPublished = firstAppearanceCommitDate gitInfo
--- a/code/hsec-tools/src/Security/Advisories/Filesystem.hs
+++ b/code/hsec-tools/src/Security/Advisories/Filesystem.hs
@@ -27,7 +27,7 @@ module Security.Advisories.Filesystem
 import Control.Applicative (liftA2)
 import Data.Bifunctor (bimap)
 import Data.Foldable (fold)
-import Data.Functor ((<&>))
+import Data.Functor (($>))
 import Data.Semigroup (Max(Max, getMax))
 import Data.Traversable (for)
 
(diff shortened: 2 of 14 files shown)


code/hsec-tools/src/Security/Advisories/Filesystem.hs
```diff
@@ -129,10 +130,12 @@ listAdvisories root =
       then return $ pure []
       else do
         oob <-
-          liftIO (getAdvisoryGitInfo advisoryPath) <&> \case
-            Left _ -> emptyOutOfBandAttributes
+          liftIO (getAdvisoryGitInfo advisoryPath) >>= \case
+            Left gitErr -> 
+              liftIO (hPutStrLn stderr ("obtaining out of band attributes failed: \n" <> explainGitError gitErr))
```
TristanCacqueray, line 135: May we return a validation error instead of a warning on the stderr?

MangoIV: chesterton's fence: why's the error ignoed in the first place, are there valid reasons to not abort here?

MangoIV:
something else I was wondering. Do we really want `ZonedTime`? It has some really awful properties like it missing an `Ord` instance. What do you think about changing everything to `UTCTime`, @TristanCacqueray?

MangoIV:
… it seems like this is an endeavour for another day, toml-parser doesn’t support it (though it should)

code/hsec-tools/src/Security/Advisories/Filesystem.hs
MangoIV: I think the idea is that we don't want to fail if there's no git information as there might still be one supplied by the advisory itself. I think this is an issue with the structure of the code.

MangoIV: I have decided to bubble the error up instead of using this `oobValue` with optional fields. I think it was just not very fortunately designed.

MangoIV:
I have refactored the out of band stuff such that we can bubble up information on why it failed. This is important if git doesn't work as expected.

MangoIV:
I can backport 4b and fe to the minor version if you don’t want breaking changes but I don’t know if it’d be too helpful.

code/hsec-tools/src/Security/Advisories/Filesystem.hs
TristanCacqueray: Thanks, that looks reasonable to me. Though I worry this is a complicated API for downstream users, perhaps we could do instead:
- Make the oob attributes explicit in the advisory document, and/or
- Publish a static version of the advisories with the oob attributes pre resolved.

What do you think @frasertweedale ?

MangoIV: Publishing a version with the oob pre resolved would be a good idea anyways. That would make testing without git possible and generally would allow distribution without git. I don’t think the interface for downstream users is more complicated now, tbh it was quite complicated with these fields that could independently be Nothing when in reality I don’t see a reason why they should be able to. Not attaching a reason for why the oob attributes being missing was problematic as well, imo, this caused the Git errors to never be shown.

telser:
I believe this will resolve an issue that currently causes https://github.com/MangoIV/cabal-audit to break for me. Though @MangoIV can correct me if that is incorrect.

MangoIV:
Yes @frasertweedale already knows :)

code/hsec-tools/app/Main.hs
```diff
@@ -153,24 +155,18 @@ withAdvisory :: (Maybe FilePath -> Advisory -> IO ()) -> Maybe FilePath -> IO ()
...
-      getAdvisoryGitInfo path <&> \case
-        Left _ -> id
-        Right gitInfo -> \oob -> oob
-          { oobPublished = Just (firstAppearanceCommitDate gitInfo)
-          , oobModified = Just (lastModificationCommitDate gitInfo)
-          }
+  oob <- runExceptT case file of
+    Nothing -> throwE StdInHasNoOOB
+    Just path -> withExceptT GitHasNoOOB do 
+      gitInfo <- ExceptT $ liftIO $ getAdvisoryGitInfo path
```
blackheaven, line 161: Isn't liftIO enough? (ExceptT should have the instance)

code/hsec-tools/src/Security/Advisories/Git.hs
```diff
@@ -117,4 +119,7 @@ getAdvisoryGitInfo path = do
       -- the same as `ExitFailure`
       pure . Left $ GitProcessError status stdout stderr
   where
-    parseTime s = maybe (Left $ GitTimeParseError s) Right $ iso8601ParseM s
+    parseTime :: String -> Either GitError ZonedTime
+    parseTime s = maybe (Left $ GitTimeParseError s) Right $
+       iso8601ParseM s
+         <|> utcToZonedTime utc <$> iso8601ParseM s
```
blackheaven, line 125: I'm not sure it does preserve the date

code/hsec-tools/src/Security/Advisories/Git.hs
```diff
@@ -117,4 +119,7 @@ getAdvisoryGitInfo path = do
       -- the same as `ExitFailure`
       pure . Left $ GitProcessError status stdout stderr
   where
-    parseTime s = maybe (Left $ GitTimeParseError s) Right $ iso8601ParseM s
+    parseTime :: String -> Either GitError ZonedTime
+    parseTime s = maybe (Left $ GitTimeParseError s) Right $
+       iso8601ParseM s
```
blackheaven, line 124: We should force the format at git level: https://stackoverflow.com/a/7651782

blackheaven (changes requested):
Quick note aside: I'm not a fan of BlockArguments

@MangoIV can you check whether forcing git date format fixes the issue please?

code/hsec-tools/app/Main.hs
MangoIV: `it's an `IO (Either _ _)`, `liftIO` give's you an `ExceptT IO (Either _ _)`, no?

code/hsec-tools/src/Security/Advisories/Git.hs
MangoIV: I tried to pass `--date=local` and it didn't work. I'm starting to think that a non-zoned time should be a parse into `UTCTime`.

MangoIV: > I'm not sure it does preserve the date

what do you mean?

code/hsec-tools/src/Security/Advisories/Git.hs
MangoIV: I tried to pass `--date=local` and it didn't work. This is in general not really ideal, we shouldn't depend on the git installation if git itself doesn't guarantee a stable interface to these things and/ or this is not documented. 

There's also https://github.com/haskell/time/issues/257 which I don't know whether it's the case.

MangoIV: https://github.com/MangoIV/cabal-audit/actions/runs/9321254777/job/25659857545

here's a run. the build step is `Test git` which uses the date format local and it still outputs a non-zoned iso8601. I wouldn't have done the separate parse try without trying that first.

MangoIV: *I assume* (read "guess") that it *is* valid to output a non-zoned iso8601 even with `local` *if* the computer is itself in `UTC` which is why I'm not sure whether parsing an unzoned time into a `UTCTime` should just work.

code/hsec-tools/src/Security/Advisories/Git.hs
TristanCacqueray: I just validated that the osv files are not changed by this PR with:

```
mkdir osvs
for i in $(find advisories/ -name "*.md" | grep -v reserved); do cabal run exe:hsec-tools -- osv $i > osvs/$(basename $i); done
```

TristanCacqueray (approved):

code/hsec-tools/app/Main.hs
blackheaven: Sorry, I thought it was only IO

code/hsec-tools/src/Security/Advisories/Git.hs
blackheaven: It's not that obvious, we get a time and day without time zone, assuming UTC seems audacious. 

IMO, we should rely only on UTCTime, enforcing it in the git command.

code/hsec-tools/src/Security/Advisories/Git.hs
blackheaven: Let's do this, it seems to be the lesser bad solution we have.

code/hsec-tools/src/Security/Advisories/Git.hs
TristanCacqueray: Oh I see, yes I'm on UTC. Then I remove my approval. Perhaps we could just parse an epoch timestamp if possible?

MangoIV: hsec-tools already does this several times (assuming UTC when there's not enough information) That's why I opened the issue about removing ZonedTime from hsec-tools. It makes stuff harder as well. It *is* however the correct *time* to assume, it's not *wrong*, it's just not semantically as nice. See https://github.com/haskell/security-advisories/issues/203 cc @frasertweedale who commented on #203

code/hsec-tools/src/Security/Advisories/Git.hs
MangoIV: I'm confused, setting it in git is a non-solution because it doesn't work. We *have* to do it like this, no?

blackheaven: I'm confused, I thought that, requiring a ZonedTime from git wasn't working, requiring a UTCTime doesn't work either?

blackheaven: @frasertweedale was talking of a flag few minutes ago (at ZuriHac), maybe we can have a try

MangoIV:
I have replaced ZonedTime with UTCTime; there's now not a single place where ZonedTime occurs anymore

blackheaven (approved):

MangoIV:
how do I check if the advisories are still valid btw? :3

MangoIV:
don't wanna break your stuff :3

MangoIV:
thank you for your review @blackheaven @TristanCacqueray  <3

blackheaven:
> how do I check if the advisories are still valid btw? :3

the CI handles it :)

MangoIV:
ah perfect; I think I need to do some version bumps before this can get merged, no?

blackheaven (approved):

blackheaven (approved):

Issue #167, opened 2024-03-18
add CVSS 4.0 support


frasertweedale:
## Summary

CVSS 4.0 is here, and already supported by OSV.  Add support for it in our cvss lib and hsec-tools.

unorsk:
Hey! I'll take this one! :)

TristanCacqueray:
@unorsk Thanks! I wrote the initial CVSS library, please let me know if you need any help.

unorsk:
Started working on this. No wonder this hasn't been implemented yet 😅
The way they changed the scoring system isn't very straightforward... but it's fun :)

TristanCacqueray:
It seems like you can find an JavaScript implementation in https://www.first.org/cvss/calculator/app.js?v=7 . Looks pretty onerous, good luck!

frasertweedale:
Might be better to just read the spec: https://www.first.org/cvss/v4.0/specification-document
and test the implementation using the official examples: https://www.first.org/cvss/v4.0/examples

PR #208, opened 2024-06-12
Cvss4.0 support

unorsk:
Fixing [add CVSS 4.0 support](https://github.com/haskell/security-advisories/issues/167)


frasertweedale:
Thanks for the updates @unorsk.  I'll try and review them this weekend.

unorsk:
> Thanks for the updates @unorsk. I'll try and review them this weekend.

@frasertweedale  You can take a look at it if you want, but this isn't ready yet :) There is one thing I commented out in the tests that I am going to fix and lots of other places in the code that need some love. One of the reasons it took me so long is that I made a rewrite of the reference implementation in TypeScript which I used as a reference for my Haskell implementation, that (not surprisingly) isn't very idiomatic.
And then there is a new metric **U**rgency that can have values 'Red / Amber / Green / Clear' 🙈 in contrast to the rest of the metrics which can only have single character values – haven't solved this one yet.

code/cvss/src/Security/CVSS40Lookup.hs
```diff
@@ -0,0 +1,325 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Security.CVSS40Lookup (lookupScore, maxComposed, maxComposedEQ3, maxSeverityeq3eq6, maxSeverity) where
```
frasertweedale, line 3: (just a code organisation thing) - I would like this module to be beneath `Security.CVSS`, e.g. `Security.CVSS.V4_0`.

unorsk: @frasertweedale done

unorsk:
@frasertweedale, it's kind of ready 🙈

frasertweedale:
> @frasertweedale, it's kind of ready 🙈

Thanks @unorsk.  I've had a quick look; I'll need to set aside some time to understand the implementation - perhaps (hopefully!) this weekend.

frasertweedale:
So, I've had a look and it's a solid start - thanks @unorsk!  I'm working on some improvements using sum types for the MicroVectors and a total function for the score lookup, rather than the maps and lookup tables.

It seems that the scoring function is underspecified in the spec doc.  There are some behaviours in the reference implementation that, from what I can see, aren't explained in the spec but rather fill in gaps or resolve ambiguities.  I might be missing something but the spec seems rather poor or at least incomplete.  Sigh...

unorsk:
> So, I've had a look and it's a solid start - thanks @unorsk! I'm working on some improvements using sum types for the MicroVectors and a total function for the score lookup, rather than the maps and lookup tables.
Yeah, sure.

> It seems that the scoring function is underspecified in the spec doc. There are some behaviours in the reference implementation that, from what I can see, aren't explained in the spec but rather fill in gaps or resolve ambiguities. I might be missing something but the spec seems rather poor or at least incomplete. Sigh...

The spec isn't great 😅

unorsk:
The implemenation wasn't great and there was some room for improvement as far as I remember 😅
But as I said - the spec could've been better and maybe I could help somehow to move the case further? Even though it's been almost a year and I already forgot some of the details 🙈

unorsk:
The implemenation wasn't great and there was some room for improvement as far as I remember 😅
But as I said - the spec could've been better and maybe I could help somehow to move the case further? Even though it's been almost a year and I already forgot some of the details 🙈

frasertweedale:
> The implemenation wasn't great and there was some room for improvement as far as I remember 😅 But as I said - the spec could've been better and maybe I could help somehow to move the case further? Even though it's been almost a year and I already forgot some of the details 🙈

Yeah, I have a WIP branch with a lot of improvements, but I ran out of capacity to complete the work.  I'll push my branch before ZuriHac in case someone wants to try and make progress.  Cheers!

PR #213, opened 2024-06-27
Add Advisory.Ecosystem to support GHC's advisory

### LINKED ISSUE #212: Handle ghc (non hackage) advisories
## Summary

To support #210 , we need to handle `advisories/ghc` paths in hsec-tools:

- decide and document the "component" names in the `ghc` namespace, e.g. `compiler`, `GHCi`, `RTS`
- update OSV generation code

TristanCacqueray:
This change adds a new OOB to pass the Ecosystem value from the advisory path.

Fixes #212

### DIFF
--- a/code/hsec-core/src/Security/Advisories/Core/Advisory.hs
+++ b/code/hsec-core/src/Security/Advisories/Core/Advisory.hs
@@ -53,7 +53,7 @@ data Ecosystem = Hackage Text | GHC GHCComponent
 
 -- Keep this list in sync with the 'ghcComponentFromText' below
 data GHCComponent = GHCCompiler | GHCi | GHCRTS
-  deriving stock (Show, Eq)
+  deriving stock (Show, Eq, Enum, Bounded)
 
 ghcComponentToText :: GHCComponent -> Text
 ghcComponentToText c = case c of
--- a/code/hsec-core/src/Security/Advisories/Core/Advisory.hs
+++ b/code/hsec-core/src/Security/Advisories/Core/Advisory.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE DerivingVia, OverloadedStrings #-}
 
 module Security.Advisories.Core.Advisory
   ( Advisory(..)
@@ -10,6 +10,10 @@ module Security.Advisories.Core.Advisory
   , AffectedVersionRange(..)
   , OS(..)
   , Keyword(..)
+  , Ecosystem(..)
+  , GHCComponent(..)
+  , ghcComponentToText
+  , ghcComponentFromText
   )
   where
 
@@ -44,10 +48,30 @@ data Advisory = Advisory
   }
   deriving stock (Show)
 
+data Ecosystem = Hackage Text | GHC GHCComponent
+  deriving stock (Show, Eq)
+
+-- Keep this list in sync with the 'ghcComponentFromText' below
+data GHCComponent = GHCCompiler | GHCi | GHCRTS
+  deriving stock (Show, Eq)
+
+ghcComponentToText :: GHCComponent -> Text
+ghcComponentToText c = case c of
+  GHCCompiler -> "compiler"
+  GHCi -> "ghci"
+  GHCRTS -> "rts"
+
+ghcComponentFromText :: Text -> Maybe GHCComponent
+ghcComponentFromText c = case c of
+  "compiler" -> Just GHCCompiler
+  "ghci" -> Just GHCi
+  "rts" -> Just GHCRTS
+  _ -> Nothing
+
 -- | An affected package (or package component).  An 'Advisory' must
 -- mention one or more packages.
 data Affected = Affected
-  { affectedPackage :: Text
+  { affectedEcosystem :: Ecosystem
   , affectedCVSS :: CVSS.CVSS
   , affectedVersions :: [AffectedVersionRange]
   , affectedArchitectures :: Maybe [Architecture]
--- a/code/hsec-core/src/Security/Advisories/Core/Advisory.hs
+++ b/code/hsec-core/src/Security/Advisories/Core/Advisory.hs
@@ -52,7 +52,7 @@ data ComponentIdentifier = Hackage Text | GHC GHCComponent
   deriving stock (Show, Eq)
 
 -- Keep this list in sync with the 'ghcComponentFromText' below
-data GHCComponent = GHCCompiler | GHCi | GHCRTS
+data GHCComponent = GHCCompiler | GHCi | GHCRTS | GHCPkg | RunGHC | IServ | HP2PS | HPC | HSC2HS | Haddock
   deriving stock (Show, Eq, Enum, Bounded)
 
 ghcComponentToText :: GHCComponent -> Text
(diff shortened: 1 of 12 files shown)


TristanCacqueray:
I am not sure what to do with the advisory schema. Presently, this assumes the GHC component will be set as the `affected.package` in the front-matter, but perhaps we should also change the schema to support `affected.component` instead? Though, we don't have the OOB when parsing the advisory...

code/hsec-core/src/Security/Advisories/Core/Advisory.hs
```diff
@@ -44,6 +47,13 @@ data Advisory = Advisory
   }
   deriving stock (Show)
 
+data Ecosystem = Hackage | GHC GHCComponent
```
blackheaven, line 50: I think it should be `Affected`, or be the type of `Affected.affectedPackage`, which might be renamed `Affected.affectedComponent`.

Otherwise we might be able to name a package (e.g. `aeson`) while being in `GHCi`.

TristanCacqueray: Though we can already mismatch a package with the name from the path. So it seems like we need an extra validation to ensure the package/ghc component from the path match at least one of the affected.

Well I assumed a given advisory would apply to a single `Ecosystem`, but perhaps this is not true. In that case, I think we need to update the affected schema to have either a `package` or a `component` key to parse a single `Affected.affectedEcosystem` and remove the `affectedPackage`.

```haskell
data AffectedEcosystem = Hackage PackageName | GHC GHCComponent
```

What do you think?

blackheaven: It seems the right way to do it, thanks.

TristanCacqueray: This is now implemented, could you please take another look?

code/hsec-core/src/Security/Advisories/Core/Advisory.hs
```diff
@@ -44,10 +48,30 @@ data Advisory = Advisory
   }
   deriving stock (Show)
 
+data Ecosystem = Hackage Text | GHC GHCComponent
```
frasertweedale, line 51: This is more than the ecosystem: it is a component/package identifier qualified (via constructor) by ecosystem.  I approve of the structure of this data type, but I would prefer to name it something else.  Some suggestions:

- `PackageIdentifier`
- `AffectedPackageIdentifier`
- `ComponentIdentifier`
- `AffectedComponentIdentifier`

I lean toward `ComponentIdentifier` (the fact that the component is *affected* by some issue is incidental to its construction).  This is because, in future, we might further refine the type to allow specifying a sub-package *component* (e.g. `hsec-tools:exe`, if there were an issue in the executable but the `lib` component was unaffected).

code/hsec-core/src/Security/Advisories/Core/Advisory.hs
```diff
@@ -44,10 +48,30 @@ data Advisory = Advisory
...
+  GHCCompiler -> "compiler"
+  GHCi -> "ghci"
+  GHCRTS -> "rts"
+
+ghcComponentFromText :: Text -> Maybe GHCComponent
+ghcComponentFromText c = case c of
+  "compiler" -> Just GHCCompiler
+  "ghci" -> Just GHCi
+  "rts" -> Just GHCRTS
+  _ -> Nothing
```
frasertweedale, line 69: This is good.  We should also file a PR to osv-schema to document these values (I will do this now).

frasertweedale: For completeness, we should probably also add:
- `ghc-pkg`
- `runghc`
- `iserv`
- `hp2ps`
- `hpc` (only old versions on hackage)
- `hsc2hs` (recent versions are on hackage but it is also in the GHC distribution so we should be able to represent that).
- `haddock` (see `hsc2hs`)

TristanCacqueray:
Thank you for the review @frasertweedale , I've updated the PR accordingly.

code/hsec-tools/src/Security/Advisories/Format.hs
```diff
@@ -135,16 +136,25 @@ instance Toml.ToTable AdvisoryMetadata where
     ["aliases"   Toml..= amdAliases x | not (null (amdAliases x))] ++
     ["related"   Toml..= amdRelated x | not (null (amdRelated x))]
 
+instance Toml.FromValue GHCComponent where
+  fromValue v = case v of
+    Toml.Text' _ n
+      | Just c <- ghcComponentFromText n -> pure c
+    _ -> Toml.failAt (Toml.valueAnn v) "Invalid component, expected compiler|ghci|rts"
```
frasertweedale, line 143: This message is does not include all the options.  It might be worthwhile to define a lookup table that we can use for parsing and also for producing a list of the valid options.

code/hsec-core/src/Security/Advisories/Core/Advisory.hs
```diff
@@ -44,10 +48,44 @@ data Advisory = Advisory
...
+  RunGHC -> "runghc"
+  IServ -> "iserv"
+  HP2PS -> "hp2ps"
+  HPC -> "hpc"
+  HSC2HS -> "hsc2hs"
+  Haddock -> "haddock"
+
+ghcComponentFromText :: Text -> Maybe GHCComponent
+ghcComponentFromText c = case c of
+  "compiler" -> Just GHCCompiler
```
frasertweedale, line 73: I wonder if we should just use the string `"ghc"` instead of `"compiler"`, so that in all relevant cases the string is just the executable name.  This would also mean using `"ghc-iserv"` instead of `"iserv"`.

The only case that does not correspond to a particular executable is then "rts".  This is fine (we do need to represent it).

frasertweedale (commented):
@TristanCacqueray thank you for the quick turnaround.  I have just a couple more small suggestions.

code/hsec-tools/src/Security/Advisories/Format.hs
TristanCacqueray: Oh well, we already have the conversion functions.

code/hsec-core/src/Security/Advisories/Core/Advisory.hs
TristanCacqueray: Good idea, this is now implemented.

code/hsec-tools/src/Security/Advisories/Queries.hs
```diff
@@ -38,9 +38,10 @@ isAffectedByHelper checkWithRange queryPackageName queryVersionish =
     any checkAffected . advisoryAffected
     where
       checkAffected :: Affected -> Bool
-      checkAffected affected =
-        queryPackageName == affectedPackage affected
-          && checkWithRange queryVersionish (fromAffected affected)
+      checkAffected affected = case affectedComponentIdentifier affected of
+        Hackage pkg -> queryPackageName == pkg && checkWithRange queryVersionish (fromAffected affected)
+        -- TODO: support GHC ecosystem query, e.g. by adding a cli flag
```
blackheaven, line 43: I'm not sure it'll be relevant

code/hsec-tools/test/Spec/FormatSpec.hs
```diff
@@ -70,13 +70,22 @@ genAdvisoryMetadata =
...
     <*> (Map.toList . Map.fromList <$> Gen.list (Range.linear 0 5) ((,) <$> genText <*> genVersionRange))
 
+genComponentIdentifier :: Gen.Gen ComponentIdentifier
+genComponentIdentifier = Gen.choice $
+  [ Hackage <$> genText
+  , GHC <$> genGHCComponent
+  ]
+
+genGHCComponent :: Gen.Gen GHCComponent
+genGHCComponent = Gen.choice $ map pure [minBound..maxBound]
```
blackheaven, line 87: I think you can use [enumBounded](https://hackage.haskell.org/package/hedgehog-1.4/docs/Hedgehog-Gen.html#v:enumBounded)

blackheaven (approved):

frasertweedale:
@TristanCacqueray I rebased this PR on latest `main` (rebase was clean) and pushed one additional commit, which updates the `GHCComponent` parsing to distinguish these error cases and return a more specific error message:
- `ghc-component` is not a TOML text value
- `ghc-comopnent` is a TOML text value, but the value is not recognised

If you are happy with these changes, then I think it's good to merge.

code/hsec-tools/test/Spec/FormatSpec.hs
TristanCacqueray: Thanks for the tip!

code/hsec-tools/src/Security/Advisories/Queries.hs
TristanCacqueray: I agree.

TristanCacqueray:
@frasertweedale looks good to me, thanks!

PR #220, opened 2024-07-18
support GHC 9.10

frasertweedale:
9.10 build fails due to bounds on `feed` library not admitting `base-4.20.0.0` (upstream issue: https://github.com/haskell-party/feed/issues/73 )

Apart from that, this PR should fix the CI issues.

*edit* converting to draft until all deps support GHC 9.10.
In the meantime, I created https://github.com/haskell/security-advisories/pull/221 which deals with the CI failures arising due to libc version conflicts with the platform image (and only that issue).


.github/workflows/haskell-ci.yml
```diff
@@ -162,29 +162,17 @@ jobs:
           rm -f cabal-plan.xz
           chmod a+x $HOME/.cabal/bin/cabal-plan
           cabal-plan --version
-      - name: install hlint
```
blackheaven, line 165: Do we drop it purposefully?

blackheaven (approved):

PR #228, opened 2024-07-30
html: fix missing <tr> on by-packages index

frasertweedale:
Presentation of packages/components that have multiple advisories is corrupted, due to missing <tr> parent elements.  Data cells for multiple advisories are appearing in a single row.  Add missing <tr> elements.

Before:

image

After:

image

---

## Advisory

- [ ] It's not duplicated
- [ ] All fields are filled

## hsec-tools


blackheaven (approved):
Does it change the display?

frasertweedale:
> Does it change the display?

Yes; I updated the PR description with before/after screenshots.

PR #214, opened 2024-07-10
 Add advisories for numeric bugs in GHC 

TristanCacqueray:
Fixes #210

PR #230, opened 2024-07-31
add advisory for biscuit-haskell 0.3.x

divarvel:
## Advisory

- [x] It's not duplicated
- [x] All fields are filled

## hsec-tools

---

Following the publication of https://github.com/biscuit-auth/biscuit-haskell/security/advisories/GHSA-47cq-pc2v-3rmp

It's my first time, let me know if I did it correctly :-)

For hsec-tools, i'm not sure how to do it (or if i should do it myself, vs having it done by CI).


blackheaven (approved):

TristanCacqueray:
Yes, CI validates it. I don't remember, what's the process to pick the HSEC id again?

blackheaven:
IIRC the merger amend the commit.

Anyway, we should document it.

/cc @frasertweedale

divarvel:
I have moved the PR back to draft because the CVE ID might change (we filed several CVEs but we might have to only keep a single CVE that covers spec and implementations). Is this something that can be amended later or does it need to be right before merging?

frasertweedale:
> I have moved the PR back to draft because the CVE ID might change (we filed several CVEs but we might have to only keep a single CVE that covers spec and implementations). Is this something that can be amended later or does it need to be right before merging?

We can merge without the CVE alias (or with the current CVE(s)), and update it later if needed.

advisories/hackage/biscuit-haskell/HSEC-0000-0000.md
```diff
@@ -0,0 +1,45 @@
+```toml
+[advisory]
+# Identifier for the advisory (mandatory). Will be assigned a "HSEC-YYYY-NNNN"
+# identifier e.g. HSEC-2022-0001. Please use "HSEC-0000-0000" in PRs.
+id = "HSEC-0000-0000"
```
frasertweedale, line 5: Please use HSEC-2024-0009 for this advisory.

advisories/hackage/biscuit-haskell/HSEC-0000-0000.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Publication date of the advisory as an RFC 3339 date.
+# DO NOT INCLUDE THIS in files committed to Git.
+# It will be derived from the Git commit history.
+# date = 2024-07-31
```
frasertweedale, line 10: Please drop these comment lines (and other comments explaining the format)

advisories/hackage/biscuit-haskell/HSEC-0000-0000.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+keywords = ["biscuit"]
+
+# Vulnerability aliases, e.g. CVE IDs (optional but recommended)
+# Request a CVE for your HSec vulns: https://iwantacve.org/
+aliases = ["CVE-2024-41949"]
```
frasertweedale, line 17: If we need to add or change the CVE-ID later, just send another PR :)

Please also add the GHSA ID to the `aliases` list.

frasertweedale:
> Yes, CI validates it. I don't remember, what's the process to pick the HSEC id again?

We have some code for working out the next unassigned/unreserved HSEC ID, but it is only used in the `reserve` command.  Still, even something like `hsec-tools next` that simply prints out the next value could be useful.  I could code that up over the weekend.

For this advisory, I proposed `HSEC-2024-0009` - after 0006..0008 which are awaiting merge in PR https://github.com/haskell/security-advisories/pull/214 (avoid conflicts).

advisories/hackage/biscuit-haskell/HSEC-0000-0000.md
divarvel: done

advisories/hackage/biscuit-haskell/HSEC-0000-0000.md
divarvel: done

advisories/hackage/biscuit-haskell/HSEC-0000-0000.md
divarvel: done, i have added both GHSA ids, the one on the spec and the one on the haskell lib.

It might be a good idea to add a mention to GHSA ids in the template comment, along with the CVE id

frasertweedale: Thanks, yes that's a good idea.

frasertweedale (approved):

frasertweedale:
Thank you for your contribution, @divarvel.

PR #268, opened 2025-03-26
advisory: Double Public Key Signing Function Oracle Attack on Ed25519

blackheaven:
---

## Advisory

- [x] It's not duplicated
- [x] All fields are filled

Discussed here: https://github.com/awakesecurity/gen-ed25-keypair/issues/8#issuecomment-2745975829

### DIFF
--- a/advisories/hackage/cryptonite/HSEC-2025-0002.md
+++ b/advisories/hackage/cryptonite/HSEC-2025-0002.md
@@ -1,0 +1,66 @@
+```toml
+[advisory]
+id = "HSEC-2025-0002"
+cwe = []
+keywords = ["crypto"]
+related = ["GHSA-w5vr-6qhr-36cc"]
+
+[[affected]]
+package = "cryptonite"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:N/I:H/A:N"
+[[affected.versions]]
+introduced = "0.1"
+
+[[affected]]
+package = "crypton"
+cvss = "CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:N/I:H/A:N"
+[[affected.versions]]
+introduced = "0.31"
+fixed = "1.0.3"
+
+[[references]]
+type = "ARTICLE"
+url = "https://portswigger.net/daily-swig/dozens-of-cryptography-libraries-vulnerable-to-private-key-theft"
+[[references]]
+type = "ARTICLE"
+url = "https://github.com/MystenLabs/ed25519-unsafe-libs"
+[[references]]
+type = "ADVISORY"
+url = "https://github.com/advisories/GHSA-w5vr-6qhr-36cc"
+[[references]]
+type = "EVIDENCE"
+url = "https://hackage.haskell.org/package/cryptonite-0.30/docs/src/Crypto.PubKey.Ed25519.html#sign"
+[[references]]
+type = "EVIDENCE"
+url = "https://github.com/haskell-crypto/cryptonite/blob/cryptonite-v0.30/cbits/ed25519/ed25519.c#53"
+[[references]]
+type = "EVIDENCE"
+url = "https://github.com/kazu-yamamoto/crypton/blob/48fb9df2de5ee752196724b081f4d3cdb57576ed/cbits/ed25519/ed25519.c#L53"
+[[references]]
+type = "FIX"
+url = "https://github.com/kazu-yamamoto/crypton/pull/47"
+
+```
+
+# Double Public Key Signing Function Oracle Attack on Ed25519
+
+The standard specification of Ed25519 message signing involves providing the
+algorithm with a message and private key.
+
+The function will use the private key to compute the public key and sign the message.
+Some libraries provide a variant of the message signing function that also takes
+the pre-computed public key as an input parameter.
+
+Libraries that allow arbitrary public keys as inputs without checking if the
+input public key corresponds to the input private key are vulnerable to the
+following attack.
+
+By using several public keys and messages, a malicious user with access to the
+signing mechanism may build up insights into the private key parameters
+resulting in access to the private key.
+
+This shortcoming means that an attacker could use the signing function as an
+Oracle, perform crypto-analysis and ultimately get at secrets.
+For example, an attacker who can’t access the private key but can access
+the signing mechanism through an API call could use several public keys and
+messages to gradually build up insights into private key parameters.
--- a/advisories/hackage/crypton/HSEC-2025-0002.md
+++ b/advisories/hackage/crypton/HSEC-2025-0002.md
@@ -1,0 +1,1 @@
+../cryptonite/HSEC-2025-0002.md


advisories/hackage/cryptonite/HSEC-2025-0002.md
```diff
@@ -0,0 +1,41 @@
...
+
+[[references]]
+type = "ARTICLE"
+url = "https://portswigger.net/daily-swig/dozens-of-cryptography-libraries-vulnerable-to-private-key-theft"
+[[references]]
+type = "ADVISORY"
+url = "https://github.com/advisories/GHSA-w5vr-6qhr-36cc"
+[[references]]
+type = "EVIDENCE"
+url = "https://hackage.haskell.org/package/cryptonite-0.30/docs/src/Crypto.PubKey.Ed25519.html#ccryptonite_ed25519_sign"
```
TristanCacqueray, line 21: This function is not exported, shouldn't this point at "#sign" ?

advisories/hackage/cryptonite/HSEC-2025-0002.md
```diff
@@ -0,0 +1,41 @@
+```toml
+[advisory]
+id = "HSEC-2025-0002"
+cwe = []
+keywords = ["crypto", "historical"]
+related = ["GHSA-w5vr-6qhr-36cc"]
+
+[[affected]]
+package = "cryptonite"
```
TristanCacqueray, line 9: It looks like this also affecting crypton (https://github.com/kazu-yamamoto/crypton/blob/main/Crypto/PubKey/Ed25519.hs#L104)

advisories/hackage/cryptonite/HSEC-2025-0002.md
```diff
@@ -0,0 +1,41 @@
...
+# Double Public Key Signing Function Oracle Attack on Ed25519
+
+The standard specification of Ed25519 message signing involves providing the
+algorithm with a message and private key.
+
+The function will use the private key to compute the public key and sign the message.
+Some libraries provide a variant of the message signing function that also takes
+the pre-computed public key as an input parameter.
+
+Some libraries were allowing arbitrary public keys as inputs without checking
```
TristanCacqueray, line 34: Perhaps this could be rephrased as "Libraries that allow arbitrary public keys as inputs without checking ... are vulnerable to the following attack"

Then below, we could use this format: "By doing [action] a [actor] may [impact] resulting in [consequence].", e.g.:

```
By using several public keys and messages, a malicious user with access to the signing mechanism
may build up insights into the private key parameters resulting in access to the private key.
```

TristanCacqueray (approved):

mihaimaruseac (approved):

advisories/hackage/cryptonite/HSEC-2025-0002.md
blackheaven: Fixed, thanks.

advisories/hackage/cryptonite/HSEC-2025-0002.md
blackheaven: Ouch, they share the same code:

* https://github.com/haskell-crypto/cryptonite/blob/cryptonite-v0.30/cbits/ed25519/ed25519.c
* https://github.com/kazu-yamamoto/crypton/blob/48fb9df2de5ee752196724b081f4d3cdb57576ed/cbits/ed25519/ed25519.c#L53

I've added crypton

advisories/hackage/cryptonite/HSEC-2025-0002.md
blackheaven: Fixed, thanks.

TristanCacqueray (approved):

TristanCacqueray:
Thinking about it some more, I wonder if the affected libraries can/will fix this issue. Perhaps this should only concern users who don't validate that the public key is valid, like the gen-ed25-keypair tool. I guess it may be acceptable for crypton* to expose dangerous functions when the caller is responsible of using them correctly.

blackheaven:
> Thinking about it some more, I wonder if the affected libraries can/will fix this issue. Perhaps this should only concern users who don't validate that the public key is valid, like the gen-ed25-keypair tool. I guess it may be acceptable for crypton* to expose dangerous functions when the caller is responsible of using them correctly.

I'm not sure we can actually prevent it, but, at least, it should be part of the documentation of exposed functions.

frasertweedale:
If we issue the advisory, please also include this article in the references: 
- https://github.com/MystenLabs/ed25519-unsafe-libs

I am unsure whether issuing an advisory against the cryptographic implementations themselves is the correct approach.
The functions are not vulnerable *per se*.  Instead, the API is susceptible to misuse / error by consuming libraries.  It is certainly the case that documentation of these functions should be improved.  Here are my thoughts / observations:

- Any identified libraries / programs that use the affected API(s) in an insecure way should have advisories (we can group them all under `HSEC-2025-0002`).
- The Ed448 implementation also seems to have been vulnerable up until *crypton-1.0.1*, where the cbits got an update and they now unconditionally rederive the public key.  See commit https://github.com/kazu-yamamoto/crypton/commit/b6bac16ce2aefbf3a688c36fa25a452e7aa5d7d0 .  We could include this info in the advisory too, or issue a separate historical advisory.
- Before merging the advisory we should liaise with *crypton* maintainers to see what they want to do.  Given Ed448 implementation rederives the public key, perhaps they can apply the same approach for Ed25519.  Alternatively, they could extend the API with safer variant(s) (suggestions below), and/or improve the documentation.

```haskell
-- | Derives public key from secret key
signSafe :: ByteArrayAccess ba => SecretKey -> ba -> Signature

-- | Checks that public key corresponds to secret key
signChecked :: ByteArrayAccess ba => SecretKey -> PublicKey -> ba -> CryptoFailable Signature
```

Ping @kazu-yamamoto - what do you think?

advisories/hackage/cryptonite/HSEC-2025-0002.md
```diff
@@ -0,0 +1,59 @@
+```toml
+[advisory]
+id = "HSEC-2025-0002"
+cwe = []
+keywords = ["crypto", "historical"]
```
frasertweedale, line 5: I think we should no include `historical` tag here, because the issue is in current releases.

kazu-yamamoto:
Thank you for informing this issue. Either approach (1) and approach (2) is fine with me.

1. Keep the backward compatibility: keep `sign` as is but add the deprecated pragma, add new functions and release a minor version.
2. Make the libary completely safe: change the definition of `safe` and release a major version.

I don't have a strong opinion on this but it seems to me that approach (2) seems better in the long run.

kazu-yamamoto:
Or what about changing  the current`sign` to ignore the parameter of public key?

frasertweedale:
> Or what about changing the current`sign` to ignore the parameter of public key?

@kazu-yamamoto yes, if you re-derive the public key, that is the safest way.  The downsize is that it causes a constant overhead for every signing operation (this is why the "optimisation" exists in the first place).  I think it's acceptable to change `sign` to ignore the public key argument.  It might be worth considering exposing the existing behaviour as a functional named, say, `signUnsafe`, with appropriate warnings, for performance-critical use cases.

kazu-yamamoto:
Please give a look at: https://github.com/kazu-yamamoto/crypton/pull/47

If OK, I will change Ed448, too.

frasertweedale:
@kazu-yamamoto thanks!  I left some comments in the PR.

advisories/hackage/cryptonite/HSEC-2025-0002.md
blackheaven: Thanks, I've dropped it

blackheaven:
Added the fixed release for crypton

TristanCacqueray (approved):

frasertweedale:
I added a couple additional references, and updated the `fixed` field to v1.0.3 (1.0.4 just fixed a minor performance regression).

Thanks for preparing the advisory.  Merging now.

Issue #252, opened 2024-11-27
How should OSV.dev consumers of Haskell Security Advisory DB records provide feedback on them?


andrewpollock:
We're seeking to shorten the feedback loop by directly connecting OSV.dev downstream consumers of Haskell vulnerabilities with the Haskell Security Advisory DB for any corrective feedback on individual records.

Could you clearly define a feedback channel (file an issue, create a PR) for downstream users to follow to be able to correct a problem with an OSV record published in the Haskell Security Advisory DB?

See https://github.com/google/osv.dev/issues/2191 for more context.

mihaimaruseac:
I think an issue in this repo should work, or directly to the mailing list, but let's first discuss this within the Haskell SRT team meeting.

m15k:
Fraser can you assign this one to me

blackheaven:
Handled in #291

> * Add [..] `model.affected.database_specific.{osv,human_link}` in OSV exports

blackheaven:
Discussed in the private mailing list.

PR #293, opened 2025-08-07
chore: regroup advisories by year

blackheaven:
---

## Advisory

- [ ] It's not duplicated
- [ ] All fields are filled

## hsec-tools

### DIFF
--- a/advisories/ghc/ghc/HSEC-2024-0007.md
+++ b/advisories/ghc/ghc/HSEC-2024-0007.md
@@ -1,1 +1,1 @@
-advisories/published/2024/0007.md
+../../published/2024/0007.md
--- a/advisories/ghc/ghc/HSEC-2024-0008.md
+++ b/advisories/ghc/ghc/HSEC-2024-0008.md
@@ -1,1 +1,1 @@
-advisories/published/2024/0008.md
+../../published/2024/0008.md
--- a/advisories/ghc/ghc/HSEC-2025-0001.md
+++ b/advisories/ghc/ghc/HSEC-2025-0001.md
@@ -1,1 +1,1 @@
-advisories/published/2025/0001.md
+../../published/2025/0001.md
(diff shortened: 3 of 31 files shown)


blackheaven:
> The symbolic links need fixing up, but otherwise I'm happy to merge this.

I think it's builtin git, here's the script I have used:

```
mkdir -p advisories/published/{2023,2024,2025}
find advisories/{ghc,hackage} -type f \
    | while read -r f; do \
    target=$(echo "$f" | sed -E 's|.*/HSEC-([0-9]+)-([0-9]+)\.md|advisories/published/\1/\2.md|') 
    git mv "$f" "$target"
    touch "$f"
done
find advisories/{ghc,hackage} -name '*.md' \
    | while read -r f; do \
    target=$(echo "$f" | sed -E 's|.*/HSEC-([0-9]+)-([0-9]+)\.md|advisories/published/\1/\2.md|') 
    ln -sf "$target" "$f"
    git add "$f"
done
```

Let me know if you have an idea to fix linking.

Thanks in advance.

frasertweedale:
> My apologies for a fly-by question from an outsider here, but I've been taking inspiration from your design here for the Julia ecosystem (in following a [great tradition](https://rust-lang.zulipchat.com/#narrow/channel/146229-wg-secure-code/topic/Lessons.20learned.3F/with/283284860) :)). Would you be willing to share a bit about the rationale for this reorg — or have you described this in some place I have yet to find? I can definitely see a number of advantages! Are the symbolic links for backwards compatibility or UX (for advisory authors or maintainers or consumers?) or some other reason?
> 
> I know this is active work, and I know describing it is also work, but I just happen to be concurrently thinking about issues exactly like this. In any case, thank you for all you've done here — this is already quite helpful!

We already supported symbolic links - because a single advisory can affect multiple packages.  But previously, the actual advisory files would reside in (one of) the package-based paths.

The main motivation for this change is to have the canonical advisory content at a location derived from the advisory ID (HSEC-YYYY-NNNN).  This lets us play more nicely with osv.dev and any other system that wants to "link back" to the advisory source without knowing anything about Haskell package/component namespaces.  We retain package-based paths but they will now all be symbolic links.

See https://github.com/haskell/security-advisories/issues/252 and https://github.com/google/osv.dev/issues/2191 for more context.

PR #318, opened 2026-04-08
advisory: HSEC-2026-0006 - Cabal deletes project source files during configure (#317)

### LINKED ISSUE #317: Cabal deletes projects files in rare circumstances
See https://github.com/haskell/cabal/commit/3a9830bbdabef2f1009a69957966b778c7c1a9ee#diff-e3cd8c042e1d9d4f3d54c4ec03a508cc2e61598aa7f88033d1cf847e5b712647R1699

and https://github.com/haskell/cabal/issues/11176

My opinion is that this could qualify as potential "Arbitrary File Deletion", although the behavior alone may not enable such an attack. There may be additional requirements to really drive arbitrary file deletion, e.g.:

- assume `removeFile` follows symlinks (which is not the case at the moment)
- an attacker could now upload such a forged package, which upon build will remove the file the symlink points to

But it's clear that cabal in general should not be removing files from the source directory.

----

Mandatory information:

* Package : cabal
* cvss: `?`
* affected versions: 2.2+ (I think)

Optional:

* cve: `?`
* keywords: `?`
* aliases: `?`
* related: `?`
* affected OSes: `?`
* affected architecture: `?`
* declarations: `?`
* Long description: `?`

blackheaven:
## Advisory

- [x] It's not duplicated
- [x] All fields are filled

### DIFF
--- a/advisories/published/2026/HSEC-2026-0006.md
+++ b/advisories/published/2026/HSEC-2026-0006.md
@@ -1,0 +1,37 @@
+```toml
+[advisory]
+id = "HSEC-2026-0006"
+cwe = [73]
+keywords = ["cabal", "file-deletion", "configure"]
+
+[[references]]
+type = "REPORT"
+url = "https://github.com/haskell/cabal/issues/11176"
+
+[[references]]
+type = "INTRODUCED"
+url = "https://github.com/haskell/cabal/commit/3a9830bbdabef2f1009a69957966b778c7c1a9ee"
+
+[[affected]]
+package = "Cabal"
+cvss = "CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:H/A:N"
+
+[[affected.versions]]
+introduced = "2.2"
+```
+
+# Cabal deletes project source files during configure
+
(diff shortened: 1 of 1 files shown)


advisories/published/2026/HSEC-2026-0006.md
```diff
@@ -0,0 +1,37 @@
...
+
+This behavior was introduced in commit `3a9830b` to resolve header precedence
+issues, as C compilers prefer relative includes over `-I` search paths. The
+workaround uses `removeFile` on source directory files, which is destructive and
+should not happen during a build process.
+
+While the current implementation does not follow symlinks, the deletion of source
+files during a build operation is a security issue. This could potentially lead to
+arbitrary file deletion if combined with symlink-following behavior in future
+versions or specific system configurations.
```
hasufell: Well, as demonstrated, we CAN delete headers outside of the project directory on windows without the need for symlinks.

blackheaven: Thanks :)

hasufell (approved):

mihaimaruseac (approved):
Thank you!

PR #326, opened 2026-05-21
advisory DoS on floating point exponent and year

blackheaven:
---

## Advisory

- [x] It's not duplicated
- [x] All fields are filled


Lysxia (approved):

mihaimaruseac (approved):

ysangkok:
Should this be amended to say that the fix is also available in aeson-2.2 now that the patch was backported? See https://www.stackage.org/diff/lts-24.56/lts-24.57

blackheaven:
Handled in #338.

PR #332, opened 2026-06-03
advisory(HSEC-2026-0008): `crypton-x509-validation` name constraint enforcement

blackheaven:
## Advisory

- [x] It's not duplicated
- [x] All fields are filled

### DIFF
--- a/advisories/published/2026/HSEC-2026-0008.md
+++ b/advisories/published/2026/HSEC-2026-0008.md
@@ -1,0 +1,66 @@
+```toml
+[advisory]
+id = "HSEC-2026-0008"
+cwe = [295]
+keywords = ["x509", "pki", "tls", "mitm", "name-constraints"]
+aliases = ["CVE-2026-9648"]
+
+[[affected]]
+package = "crypton-x509-validation"
+cvss = "CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:H/I:H/A:N"
+
+[[affected.versions]]
+introduced = "1.6.12"
+fixed = "1.9.1"
+
+[[affected]]
+package = "crypton-x509"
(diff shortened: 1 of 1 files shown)


TristanCacqueray (approved):

ysangkok:
I just noticed that since we were planning to have crypton < 1.1 in Stackage LTS 25, Stackage users would be affected by this even though they used the latest LTS. We're tracking this upgrade here:

* https://github.com/commercialhaskell/stackage/issues/7929

If a fix is not backported, the [release tasks page for Fedora 25](https://fedorapeople.org/groups/schedule/f-45/f-45-releng-tasks.html) dictates that crypton-x509 would have to be removed from Fedora on the 29th of this month...

CC @kazu-yamamoto

juhp:
> I just noticed that since we were planning to have crypton < 1.1 in Stackage LTS 25, Stackage users would be affected by this even though they used the latest LTS. We're tracking this upgrade here:
> 
>     * [crypton-1.1 & tls-2.3 commercialhaskell/stackage#7929](https://github.com/commercialhaskell/stackage/issues/7929)

I agree it would be good to fix this for lts-25 - otherwise it might have to wait for another ~6 months in lts-26 say.

> If a fix is not backported, the [release tasks page for Fedora 25](https://fedorapeople.org/groups/schedule/f-45/f-45-releng-tasks.html) dictates that crypton-x509 would have to be removed from Fedora on the 29th of this month...

I don't think that is actually the case - perhaps this applies to Fedora packages with longstanding vulnerabilities: in particular I don't think any Fedora Security bug has been filed yet for this either. Regardless it would be good to address this in Stackage early, I feel.

kazu-yamamoto:
Since the vulnerabilities in `tls` and `crypton-x509-validation` were pointed out (by AI), I have fixed them. But they're not realistic scenarios, so I don't think there's much to be afraid of.

PR #342, opened 2026-09-23
HSEC-2026-0010: `jwt` misparsing fractional dates

ulidtko:
Hi again, HSRT.

This one scores at CVSS **4.5**. Reporting in public, because:
 * the only current maintainer had not addressed [the proposed patch][fix] (nor other patches), for over 180 days;
 * the original [report][bug] stands open since 2023;
 * and because the impact is, at best, moderate.

Please review & merge.

[bug]: https://github.com/puffnfresh/haskell-jwt/issues/4
[fix]: https://github.com/puffnfresh/haskell-jwt/pull/10

## Advisory

- [x] It's not duplicated
- [x] All fields are filled
  - 🚫 I could not install hsec-tools — the version in main fails build with:
     ```
     Error: [Cabal-7125]
     Failed to download zip-archive-0.4.3.2 (which is required by exe:hsec-tools from hsec-tools-0.5.0.0). The exception was:
       Unexpected response 403 for http://objects-us-east-1.dream.io/hackage-mirror/package/zip-archive-0.4.3.2.tar.gz
     ```


TristanCacqueray:
Thank you for the report @ulidtko , would be nice to get a reply from the maintainer. @puffnfresh

Perhaps the description could explain a bit more what is the vulnerability, e.g. shouldn't it be https://cwe.mitre.org/data/definitions/613.html ?

mihaimaruseac (approved):

ulidtko:
> shouldn't it be https://cwe.mitre.org/data/definitions/613.html

Well, the core bug as I see it, is [CWE-681](https://cwe.mitre.org/data/definitions/681.html) `Incorrect Conversion between Numeric Types`.

But in-context, I agree [CWE-613](https://cwe.mitre.org/data/definitions/613.html) `Insufficient Session Expiration` can definitely arise *in callers* to jwt lib.

Since `adrisory.cwe` is a list, I'll add both.

